PR #2, opened 2017-09-20
feat(i18n): Adds German translation

asciidisco:


emilsvennesson:
@asciidisco Thanks a lot! :-D

PR #4, opened 2017-09-30
feat(i18n): Update German translation

asciidisco:


emilsvennesson:
@asciidisco Thank you! :-D

Issue #1, opened 2017-09-13
Automatic Widevine binary installation on ARM based devices


emilsvennesson:
It would be /very/ nice if it was possible to automatically install Widevine on ARM based devices as well, but I have yet to find an official source for the binary on this architecture. I've only come across this binary on third-party hosts, which obviously is a big no as that violates the license.

@asciidisco @peak3d @trummerjo Perhaps you could shed some light on this?

trummerjo:
As far as i know is the only source for arm builds ChromeOS

asciidisco:
@emilsvennesson Google doesn't provide a CDN hosted version of libwidevine for ARM systems.
As @trummerjo said, you can only get it when downloading & extracting an Chrome OS image.

Here´s a [list of Images](https://15056223545614405154.googlegroups.com/attach/2d378a32f465b8/recovery.conf?part=0.1&view=1&vt=ANaJVrFHQuBkT5zEj3FZ4NJ6mpuN26R7lCLKSxTOD7I4TP-asoSpzEKUnQaJD29uFLRNiUUTenp7gsc39cRVf06cA0iQn-DP2x-G0miqUuzPyoWH83H17Ck)

You´d need around 1.5 GB free diskspace fro them & download is roughly 700mb (depends on the image). If you don't want to violate the license, you´d need to symlink the library from the extracted image, as copying is not allowed.

peak3d:
Maybe it could make sense to investigate the chromiumos build system (ninja) how libwidevine is pulled there

emilsvennesson:
@asciidisco @trummerjo Well, that is... unfortunate. It's kind of a hassle to mount stuff within Python and that would mean it'd have to rely on subprocess. Seeing as copying is not allowed, does that also mean that we'd have to keep the extracted image? 1.5 GB is quite a lot for people who are using a small SD card for their Raspberry Pi. Does this rule apply to the CDN hosted versions as well?

@peak3d I did a brief investigation on this, and it seems like libwidevine is not even supported in Chromium OS. :-/

Thanks for getting back to me! :-)

peak3d:
Read about cloudfront wich seems to be lightweight chormeos + widevine

emilsvennesson:
Cloudfront? Would you mind hooking me up with a link? I tried Google but couldn't find anything.

peak3d:
[...]

PR #10, opened 2018-01-30
fix for arm architecture detection

Maven85:


emilsvennesson:
Hi and thanks for the PR! Could you please elaborate what this fixes and what's wrong with the current code?

Maven85:
The current code works if arch is e.g. 'armv8b'.
But when arch is e.g. 'armv8' the code returns 'armv' for arch which is not in config.ARCH_MAP.

My changes fix that.

gismo112:
Hi @emilsvennesson , hi @Maven85 

here a Log from a Kodinerds-Forum-User with Osmc.

`10:00:30.124 T:2618454784   DEBUG: [script.module.inputstreamhelper-0.3.1]: Platform information: ('Linux', 'osmc', '3.14.29-54-osmc', '#1 SMP Thu Jan 11 17:19:56 UTC 2018', 'armv7', '')
10:00:30.124 T:2618454784   DEBUG: [script.module.inputstreamhelper-0.3.1]: jsonrpc payload: {'params': {'addonid': 'inputstream.adaptive'}, 'jsonrpc': '2.0', 'id': 1, 'method': 'Addons.GetAddonDetails'}
10:00:30.125 T:2618454784   DEBUG: [script.module.inputstreamhelper-0.3.1]: jsonrpc response: {"id":1,"jsonrpc":"2.0","result":{"addon":{"addonid":"inputstream.adaptive","type":"kodi.inputstream"}}}
10:00:30.126 T:2618454784   DEBUG: [script.module.inputstreamhelper-0.3.1]: jsonrpc payload: {'params': {'properties': ['enabled'], 'addonid': 'inputstream.adaptive'}, 'jsonrpc': '2.0', 'id': 1, 'method': 'Addons.GetAddonDetails'}
10:00:30.127 T:2618454784   DEBUG: [script.module.inputstreamhelper-0.3.1]: jsonrpc response: {"id":1,"jsonrpc":"2.0","result":{"addon":{"addonid":"inputstream.adaptive","enabled":true,"type":"kodi.inputstream"}}}
10:00:30.127 T:2618454784   DEBUG: [script.module.inputstreamhelper-0.3.1]: inputstream.adaptive is installed and enabled.
10:00:30.127 T:2618454784   DEBUG: [script.module.inputstreamhelper-0.3.1]: Unsupported Widevine architecture found: armv`

I think, is the same Problem :)

[https://paste.ubuntu.com/26489041/](url)

emilsvennesson:
@Maven85 @gismo112 Ah, so OSMC incorrectly reports its architecture as  `armv7`. I merged the PR since it doesn't hurt, but honestly I think this is something the OSMC devs should take care of.

@Maven85 I made some slight changes to prevent the function from returning `armv` if the regex search would fail somehow. Thanks again for the PR!

samnazarko:
Thanks to those who contributed.

Which devices are experiencing problems with this?

There was an issue with Vero 4K and Netflix in the past. This is caused by us having an Aarch64 kernel with 32-bit userland. I use this patch, but *now only* for Vero 4K, and not Raspberry Pi, which allows us to fixup uname: https://github.com/osmc/osmc/commit/41f5537b45985d1bbabe707c249c6bfdf1250be6

It was included in 2017.12-1 for all ARM platforms.
This will be fixed in the next update.

We only tested Netflix, which seemed to work on Vero 4K after the above commit. If anyone has a better suggestion for what uname should return, do let me know.

emilsvennesson:
@samnazarko I only have remote access to a Raspberry Pi running OSMC so I 'm unable to see if it still applies to the latest update, but it appears from the log that @gismo112 provided that OSMC returns `armv7` rather than the standard `armv7l`:

```DEBUG: [script.module.inputstreamhelper-0.3.1]: Platform information: ('Linux', 'osmc', '3.14.29-54-osmc', '#1 SMP Thu Jan 11 17:19:56 UTC 2018', 'armv7', '')```

And this is the output when running ```platform.uname()``` from the terminal:
```python
platform.uname()
('Linux', 'osmc', '4.9.29-13-osmc', '#1 SMP PREEMPT Thu Dec 21 17:18:07 UTC 2017', 'armv7l', '')
```

This previously caused this script to fail because it got rid of the last character, which in OSMC:s case returns ```armv``` rather than ```armv7``` as intended.

samnazarko:
Thanks for the response. 

The log above is taken from a Vero 4K (3.14 kernel). But I can see platform.uname has been called on a Pi (different device). 

I patched Kodi's uname to fix a problem with Netflix not working. See https://github.com/osmc/osmc/commit/41f5537b45985d1bbabe707c249c6bfdf1250be6. Vero 4K has an Aarch64 kernel but 32-bit (ARMv7) userland. This causes problems when uname is used to assess machine architecture. 

I'll see if this needs revisiting. For now, keeping the workaround in is appreciated.

PR #13, opened 2018-01-31
if kernel has block device loop built as module, we have to load it f…

mkreisl:
…irst

### DIFF
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -212,6 +212,15 @@ class Helper(object):
         else:
             return success
 
+    def _load_loop_module(self):
+        """Load loop module."""
+        cmd = ['modprobe', '-q', 'loop']
+        success = self._run_cmd(cmd, sudo=True, ask=True)
+        if success:
+            return True
+        else:
+            return False
+
     def _set_loop_dev(self):
         """Set an unused loop device that's available for use."""
         cmd = ['losetup', '-f']
@@ -227,7 +236,7 @@ class Helper(object):
     def _losetup(self, bin_path):
         """Setup Chrome OS loop device."""
         cmd = ['losetup', self._loop_dev, bin_path, '-o', self._parse_chromeos_offset(bin_path)]
-        success = self._run_cmd(cmd, sudo=True, ask=True)
+        success = self._run_cmd(cmd, sudo=True, ask=False)
         if success:
             self._attached_loop_dev = True
             return True
@@ -482,8 +491,8 @@ class Helper(object):
 
                 success = [
                     self._unzip(self._temp_path(), bin_filename),
-                    self._set_loop_dev(), self._losetup(bin_path),
-                    self._mnt_loop_dev()
+                    self._load_loop_module(), self._set_loop_dev(),
+                    self._losetup(bin_path), self._mnt_loop_dev()
                 ]
                 if all(success):
                     self._extract_widevine_from_img()
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -216,10 +216,7 @@ class Helper(object):
         """Load loop module."""
         cmd = ['modprobe', '-q', 'loop']
         success = self._run_cmd(cmd, sudo=True, ask=True)
-        if success:
-            return True
-        else:
-            return False
+        return success
 
     def _set_loop_dev(self):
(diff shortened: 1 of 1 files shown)


lib/inputstreamhelper.py
```diff
@@ -212,6 +212,15 @@ def _run_cmd(self, cmd, sudo=False, ask=True, output=True):
         else:
             return success
 
+    def _load_loop_module(self):
+        """Load loop module."""
+        cmd = ['modprobe', '-q', 'loop']
+        success = self._run_cmd(cmd, sudo=True, ask=True)
+        if success:
+            return True
```
emilsvennesson: The if/else here is superfluous. You can just return `success` directly.

emilsvennesson:
Hi and thank you for the PR!

I think we need to check if `loop` is a module. Right now, you're just blindly attempting to load it even if it's built-in already. Also, language string `30030` should be changed accordingly to inform the user that `modprobe` has to be run as root as well.

Perhaps it would be wise to unload the module again after the install? This should be a simple to implement in `_cleanup()`.

lib/inputstreamhelper.py
mkreisl: Yeah, I was thinking about that, but decided to keep this program structure

mkreisl:
We do not need to check this, because modprobe ignores this silently:
```
[19:46:26][root@kmcubie:/var/log]$ modinfo loop
modinfo: ERROR: Module loop not found.
[20:30:26][root@kmcubie:/var/log]$ modprobe loop
[20:30:36][root@kmcubie:/var/log]$ modprobe -q loop && echo success
success
[20:30:56][root@kmcubie:/var/log]$ 
```

This is output if loop is compiled into kernel

mkreisl:
> Perhaps it would be wise to unload the module again after the install? This should be a simple to implement in _cleanup().

IMO we could leave this module loaded :smile:

mkreisl:
> I think we need to check if loop is a module. Right now, you're just blindly attempting to load it even if it's built-in already. Also, language string 30030 should be changed accordingly to inform the user that modprobe has to be run as root as well.

Maybe, but `losetup` requires root rights as well and this **not** mentioned in string 30030. I checked this before PR

lib/inputstreamhelper.py
mkreisl: Removed that superfluous code

emilsvennesson:
> IMO we could leave this module loaded

I definitely think we should leave the system the way it was prior to the installation. It doesn't make any sense to keep the module loaded because we are not using it.
> Maybe, but losetup requires root rights as well and this not mentioned in string 30030. I checked this before PR

Yeah, that's a mistake on my part. We should inform the user which commands that require root access.

mkreisl:
> I definitely think we should leave the system the way it was prior to the installation. It doesn't make any sense to keep the module loaded because we are not using it.

I don't think so, it makes things complicated:
You do not know if another task or user is modprobing loop while we're executing downloading,  uncompressing ... tasks, so we do not know if it is save to remove loop module at the end or not

That's my opinion. Again - Keep It Simple Stupid

emilsvennesson:
Fair point, although the losetup setup is done after downloading and uncompressing so it shouldn't have to be loaded for more than a second or two. We could also ask the user to (manually) modprobe the module prior to the installation. 

I just don't think leaving loaded kernel modules in the system is a good approach.

mkreisl:
> I just don't think leaving loaded kernel modules in the system is a good approach.

I think we should stop discussion about that point now, because it seems to be a never ending story (you're mounting ext2 image, so if kernel has ext2/ext4 compiled as module, you have to check this too ...)

emilsvennesson:
@mkreisl Okay, we should atleast display a dialog notifying the user that `loop` will remain loaded in the system.

mkreisl:
> Okay, we should atleast display a dialog notifying the user that loop will remain loaded in the system.

It's your project.
If you think this would be useful for the user you can pop up a window (or better a toast message which automatically disappears), but IMO this is useless because most users will not care about this.
It's more important to add debug message into log if something does not work, and unfortunately this is still missing:

```
_run_cmd ...
...
        except subprocess.CalledProcessError as error:
           # here is a message missing which cmd was failing
            self._log('cmd failed with output: {0}'.format(error.output))
```

PR #16, opened 2018-02-27
disk space calculation should be on the tmp folder

dawez:
Some raspberry pi installation have a really small storage space which is not enough when trying to download ~2GB file [when downloading the chromeos image]. To overcome this problem, the tmp folder inside the addon folder, can be mounted to a external disk storage [with a ln softlink]. Therefore the disk space calculation should be based on the tmp folder.


emilsvennesson:
That makes sense. Thank you for the patch!

PR #17, opened 2018-03-24
Fix python3 crash

MrKrabat:
Hello,

as you may know, Kodi 19 will only support python3 and new addons are forced to support python2 and python3.
Currently addons using inputstreamhelper are crashing just for importing inputstreamhelper.

To fix this, this small change is required. This syntax is allowed by both python2 and python3 for octal numbers.

I did not test the Widevine download functionality with python3 but everything else seems to work.

### DIFF
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -610,7 +610,7 @@ class Helper(object):
         if self._cmd_exists('ldd'):
             if not os.access(self._widevine_path(), os.X_OK):
                 self._log('Changing {0} permissions to 744.'.format(self._widevine_path()))
-                os.chmod(self._widevine_path(), 0744)  # this needs to be octal in python 3
+                os.chmod(self._widevine_path(), 0o744)
 
             missing_libs = []
             cmd = ['ldd', self._widevine_path()]


emilsvennesson:
Hi @MrKrabat and thank you for the pull request!

Kodi is using different versions of the Python 2 interpreter depending on which OS you have. I'm not entirely sure if octal numbers in `os.chmod` is supported in all versions, so that's why I decided to play it safe.

I'm aware of the switch to Python 3 in the future and I can hopefully start working on that soon. There's probably more than the octal syntax that needs to be revised. The problem right now is that it's difficult getting hold of test builds with the Python 3 interpreter for various platforms. There is quite a lot of platform specific code in the Widevine download functionatily that needs to be tested thoroughly, and I've only seen pre-compiled builds for Windows at this point.

MrKrabat:
I can understand your concern,
another option would be to use the constants defined in stat module.
```python
import os, stat
os.chmod(self._widevine_path(), stat.S_IXUSR) # give file owner execute permissions
```
The documentation specifies chmod will only work on unix systems. For windows systems you can only set read-only

There are testbuilds for android, OSX and windows, just make sure to pick a download with "feature_python3" in name.
The OSX build should work on linux too, I guess.
http://mirrors.xbmc.org/test-builds/

PR #18, opened 2018-04-02
Possibility to disable helper

horstle:
Hi,

there was a discussion recently (in german, see [here](https://www.kodinerds.net/index.php/Thread/44211-Release-Amazon-Prime-Instant-Video/?pageNo=276)) that this helper is interfering with developer's workflows, when they test different versions of widevine.

So I thought I'd try adding an option to "disable" it. I made it, so the helper would ask wether to disable it or not only once (or as long as it's unset), with a second question when disabling it as reinforcement. Currently it can only be enabled again by manually changing/removing the value in the settings.xml. Additionally I fixed some typos in the german strings.

 I'm new to python and programming in general, but I tested it a while and it seems to work as intended.


horstle:
Any news? Did you and peak3d find a solution? Should I close this?

emilsvennesson:
@horstle Hi!

Sorry, I've been lacking time to review this. I like the idea and will hopefully have time to look into this PR more closely this weekend. :-)

emilsvennesson:
I decided to take a slightly different approach and got rid of the dialog option. This is mostly intended for developers, and I feel like it could case more problems than it solves letting regular users disabling it using a dialog because there is no simple way to revert it.

Thank you for the PR and sorry for taking so long to review it! :-)

horstle:
Alright, I was thinking about making the setting visible, so there would be a GUI way to enable it again, but it would still be relatively hard to find, so I guess this is the best solution.

No worries about taking your time for reviewing. I think it's best to make sure everything is ok before merging, especially if the code is from me! :P

Issue #20, opened 2018-04-17
[Feature Request] Custom path for tmp


Gaet81:
Hello,

Will it be possible to set a custom path to the tmp folder?

On RPi the installation request 1.9gb so it's quiet extensive when you are on a SD card.
I've tried to make a symbolic link to an external drive but the plugin use rmtree in the clean function and rmtree doesn't work on symbolic link...

Gaet81

horstle:
Hi,

I created an option to do so here:

https://github.com/horstle/script.module.inputstreamhelper/tree/custom_path

If you want to you can try it out. I just tested it and found no errors. It asks for a directory if there is not enough space.

If it works for you and @emilsvennesson  is fine with it, I'll create a PR.

emilsvennesson:
Hi!

This is definitely a good idea and would be a nice enhancement. 

@horstle A PR for this would be nice indeed. :-)

Gaet81:
Hello,

I have finally some time to give you feedback.
I've tested on 3 differents harddrive and it hasn't worked...
It seems that losetup is failing but don't know why...I've made a which and the losetup cmd exists 

which losetup
/usr/bin/losetup

The debug log is here:
https://framabin.org/?01370dfcdb2cff38#ep+48GPObOzIP5wRIlCorMG0xkdL4hVYREMOFyByNd8=

horstle:
There seems to be a problem with the sudo command. Did you change anything else on your system?

emilsvennesson:
@Gaet81 The script needs to be able to run sudo without entering a password which is the default in most Raspberry Pi distros. That doesn't seem to be the case in Arch Linux though. :open_mouth:

Gaet81:
really? how secure is that? Anyway I suppose I will need to make a viduso and update it then.
I'm not at home but as soon as I'm back I'm testing this.
________________________________
De : emilsvennesson <notifications@github.com>
Envoyé : mercredi 23 mai 2018 20:45
À : emilsvennesson/script.module.inputstreamhelper
Cc : Gaet81; Mention
Objet : Re: [emilsvennesson/script.module.inputstreamhelper] [Feature Request] Custom path for tmp (#20)


@Gaet81<https://github.com/Gaet81> The script needs to be able to run sudo without entering a password which is the default in most Raspberry Pi distros. That doesn't seem to be the case in Arch Linux though. 😮

horstle:
[...]

PR #21, opened 2018-04-20
Custom path for ARM devices with limited space

horstle:
#20

### DIFF
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -136,10 +136,7 @@ class Helper:
     @classmethod
     def _temp_path(cls):
         ''' Return temporary path, usually ~/.kodi/userdata/addon_data/script.module.inputstreamhelper/temp '''
-        temp_path = ADDON.getSetting('temp_dir')
-        if not temp_path.startswith('special://profile/addon_data/script.module.inputstreamhelper/temp'):
-            temp_path = os.path.join(temp_path, 'temp')
-        temp_path = xbmc.translatePath(temp_path)
+        temp_path = xbmc.translatePath(os.path.join(ADDON.getSetting('temp_dir'), 'temp'))
         if not xbmcvfs.exists(temp_path):
             xbmcvfs.mkdir(temp_path)
 
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -135,8 +135,11 @@ class Helper:
 
     @classmethod
     def _temp_path(cls):
-        ''' Return temporary path, usually ~/.kodi/userdata/addon_data/tmp '''
-        temp_path = os.path.join(ADDON_PROFILE, 'tmp')
+        ''' Return temporary path, usually ~/.kodi/userdata/addon_data/script.module.inputstreamhelper/temp '''
+        temp_path = ADDON.getSetting('temp_dir')
+        if not temp_path.startswith('special://profile/addon_data/script.module.inputstreamhelper/temp'):
+            temp_path = os.path.join(temp_path, 'temp')
+        temp_path = xbmc.translatePath(temp_path)
         if not xbmcvfs.exists(temp_path):
             xbmcvfs.mkdir(temp_path)
 
@@ -144,7 +147,7 @@ class Helper:
 
     @classmethod
     def _mnt_path(cls):
-        ''' Return mount path, usually ~/.kodi/userdata/addon_data/tmp/mnt '''
+        ''' Return mount path, usually ~/.kodi/userdata/addon_data/script.module.inputstreamhelper/temp/mnt '''
         mnt_path = os.path.join(cls._temp_path(), 'mnt')
         if not xbmcvfs.exists(mnt_path):
             xbmcvfs.mkdir(mnt_path)
@@ -228,6 +231,13 @@ class Helper:
         """Check whether cmd exists on system."""
         # https://stackoverflow.com/questions/377017/test-if-executable-exists-in-python
         return subprocess.call('type ' + cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE) == 0
+
+    def _update_temp_path(self, new_temp_path):
+        """"Updates temp_path and merges files."""
+        old_temp_path = self._temp_path()
+
+        ADDON.setSetting('temp_dir', new_temp_path)
+        shutil.move(old_temp_path, new_temp_path)
 
     def _helper_disabled(self):
         """Return if inputstreamhelper has been disabled in settings.xml."""
@@ -640,10 +650,14 @@ class Helper:
                 xbmcgui.Dialog().ok(localize(30004), localize(30019, os=system_os()))
                 return False
 
-            if required_diskspace >= self._diskspace():
-                xbmcgui.Dialog().ok(localize(30004),  # Not enough free disk space
-                                    localize(30018, diskspace=self._sizeof_fmt(required_diskspace)))
-                return False
+            while required_diskspace >= self._diskspace():
+                if xbmcgui.Dialog().yesno(localize(30004), localize(30055)):  # not enough space, alternative path?
+                    self._update_temp_path(xbmcgui.Dialog().browseSingle(3, localize(30909), 'files'))  # temporary path
+                    continue
+                else:
+                    xbmcgui.Dialog().ok(localize(30004),  # Not enough free disk space
+                                        localize(30018, diskspace=self._sizeof_fmt(required_diskspace)))
+                    return False
 
             if not self._cmd_exists('fdisk') and not self._cmd_exists('parted'):
                 xbmcgui.Dialog().ok(localize(30004), localize(30020, command1='fdisk', command2='parted'))  # Commands are missing
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -136,7 +136,7 @@ class Helper:
     @classmethod
     def _temp_path(cls):
         ''' Return temporary path, usually ~/.kodi/userdata/addon_data/script.module.inputstreamhelper/temp '''
-        temp_path = xbmc.translatePath(os.path.join(ADDON.getSetting('temp_dir'), 'temp'))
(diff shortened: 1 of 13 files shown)


emilsvennesson:
I will review this before the week is over... sorry for taking so long.

dagwieers:
Can you please rebase?

horstle:
Done.

horstle:
The lint issues in Travis are no issues of this PR.
The first one (no-member) is just wrong, xbmcgui.Dialog().browseSingle() does exist.
The second is simply not created by this PR, as it doesn't change the corresponding lines.

horstle:
Alright, all `pylint` issues are gone, but the `too-many-lines` and `too-many-statements` (at `_install_widevine_arm()`) are simply suppressed and should be addressed in the future.

The first argument of `xbmcgui.Dialog().browseSingle()` is actually called "type" in the documentation, but since "type" already exists, I had to rename it (to "atype"). I don't think this could create an issue, because from what I understand if somebody ever called `browseSingle()` with a keyword-argument `type=` he would redefine the builtin.

horstle:
Please review @emilsvennesson @dagwieers @mediaminister .

dagwieers:
@horstle I don't know if you are still interested to implement this. If you aren't, I understand.
At least the codebase is stable again and ready for more functionality. :wink:

horstle:
I sure am interested. Sadly, I'm not sure how much time I've got in the coming weeks, but rebasing this should be doable, probably on sunday.

horstle:
Rebase is done, one check passed. Still no idea about codecov.

dagwieers:
The patch-feedback of codecov is useless currently. Don't worry about it.
(In this case it means we are only testing 35.7% of the code in your patch)

lib/inputstreamhelper/__init__.py
```diff
@@ -133,7 +133,11 @@ def _diskspace(cls):
     @classmethod
     def _temp_path(cls):
         ''' Return temporary path, usually ~/.kodi/userdata/addon_data/tmp '''
-        temp_path = os.path.join(ADDON_PROFILE, 'tmp')
+        cust_temp_dir = ADDON.getSetting('custom_temp_dir')
+        if xbmcvfs.exists(cust_temp_dir):
+            temp_path = os.path.join(cust_temp_dir, 'tmp')
+        else:
+            temp_path = os.path.join(ADDON_PROFILE, 'tmp')
```
dagwieers: Should we not just create the directories, rather than silently not using it.

horstle: this is meant for if a removable device is set, but no longer there.

mediaminister: @dagwieers The user selected the custom temporary directory from the GUI, it definitely exists. The standard temporary directory is created if it doesn't exist. I don't see the problem here.

resources/language/resource.language.en_gb/strings.po
```diff
@@ -297,3 +300,6 @@ msgctxt "#30907"
 msgid "Remove Widevine CDM library..."
 msgstr ""
 
+msgctxt "#30909"
+msgid "Custom Path"
```
mediaminister: Please use `Custom temporary directory` consistently when you interact with the user.

resources/language/resource.language.en_gb/strings.po
```diff
@@ -225,6 +225,9 @@ msgctxt "#30054"
 msgid "disabled"
 msgstr ""
 
+msgctxt "#30055"
+msgid "You do not have enough free disk space to install [B]Widevine CDM[/B] or your specified custom path does not exist anymore or hasn't got enough disk space either. Would you like to specify a (different) custom path to temporarily use for the Chrome OS Recovery Image (e.g. on a USB)?"
```
mediaminister: Do not repeat things twice, try to be concise and clear when interacting with the user.
For instance: `There is not enough free disk space to install [B]Widevine CDM[/B] or your custom temporary directory doesn't exist anymore...`

If the user should take action, provide a detailed instruction. For instance: `Please free up at least ... gigabytes of disk space before continuing the installation.`

I recommended splitting up the disk space problem and the custom temp dir not found problem. Combining these two problems in one message makes it difficult for the user to understand which actions are needed.

mediaminister:
@horstle While testing  your pull request I came to the conclusion that there is no option to use the default temporary directory again when a custom directory was set once.
Setting the default temp directory as default in the settings.xml fixes this, so there is no need any more to call this setting "custom". We can rely completely on settings.xml to define the temp directory.

I'll add a commit to this pull request which implements this.

mediaminister:
This should be ready to merge now. Please test this first on ARM devices.

lib/inputstreamhelper/__init__.py
```diff
@@ -132,16 +132,19 @@ def _diskspace(cls):
 
     @classmethod
     def _temp_path(cls):
-        ''' Return temporary path, usually ~/.kodi/userdata/addon_data/tmp '''
-        temp_path = os.path.join(ADDON_PROFILE, 'tmp')
+        ''' Return temporary path, usually ~/.kodi/userdata/addon_data/script.module.inputstreamhelper/temp '''
+        temp_path = ADDON.getSetting('temp_dir')
+        if not temp_path.startswith('special://profile/addon_data/script.module.inputstreamhelper/temp'):
```
horstle: I'd like to propose one tiny change here:
If the default setting was changed to the path without "temp" at the end, this condition would not be necessary, but rather "temp" always added to the specified path (whether default or custom).

horstle: I created a new commit. You can revert it if you see a reason.

resources/settings.xml
```diff
@@ -5,9 +5,10 @@
     <setting id="chromeos_version" value="" visible="false"/>
     <category label="30900"> <!-- Troubleshooting -->
         <setting label="30901" type="action" id="info" action="RunScript(script.module.inputstreamhelper, info)"/>
-        <setting label="30903" help="30904" type="bool" id="disabled" value="false"/>
+        <setting label="30903" help="30904" type="bool" id="disabled" value="false" default="false"/>
```
dagwieers: I think the `value="false"` is wrong, it should be replaced with `default="false"`.

resources/settings.xml
```diff
@@ -5,9 +5,10 @@
...
         <setting label="30904" type="text" enable="false"/> <!-- disabled_warning -->
         <setting label="30905" help="30906" type="action" id="install_widevine" action="RunScript(script.module.inputstreamhelper, widevine_install)" enable="eq(-2,false)" visible="!system.platform.android"/>
         <setting label="30907" help="30908" type="action" id="remove_widevine" action="RunScript(script.module.inputstreamhelper, widevine_remove)" enable="eq(-3,false)" visible="!system.platform.android"/>
+        <setting label="30909" type="folder" id="temp_dir" source="" option="writeable" default="special://profile/addon_data/script.module.inputstreamhelper"/>
```
dagwieers: This option should really only be visible on ARM. There is no need to modify the path on other platforms IMO.

test/xbmc.py
```diff
@@ -64,8 +64,10 @@ def translatePath(path):
     ''' A stub implementation of the xbmc translatePath() function '''
     if path.startswith('special://home'):
         return path.replace('special://home', os.path.join(os.getcwd(), 'test/'))
+    if path.startswith('special://profile/addon_data/script.module.inputstreamhelper/temp'):
```
dagwieers: I think it is unfortunate that there is no special path for an addon's userdata directory, something like: `special://addon_data/temp`. But I guess it should have been a method for `xbmcaddon` instead :-/

This ought to be a regular expression so that it works for other add-ons as well. But no worries, I will take care of it when backporting.

test/xbmcgui.py
```diff
@@ -44,6 +44,12 @@ def textviewer(heading, text=None, usemono=None):
         text = kodi_to_ansi(text)
         print('\033[37;100mTEXTVIEWER:\033[35;0m [%s]\n\033[35;0m%s\033[39;0m' % (heading, text))
 
+    @staticmethod
+    def browseSingle(atype, heading, shares, mask=None, useThumbs=None, treatAsFolder=None, default=None):
+        ''' A stub implementation for the xbmcgui Dialog class browseSingle() method '''
+        print('[37;100mBROWSESINGLE:[35;0m [%s] [35;0m%s[0m' % (atype, heading))
+        return '~/.kodi/userdata/addon_data/script.module.inputstreamhelper'
```
dagwieers: I would probably return a special path (the default one).

resources/settings.xml
dagwieers: Why not call this `temp_path` instead? We use that internally as well.

dagwieers:
I think I fixed the coverage reporting now. Thanks for your patience!
If you would rebase your branch (`git pull --rebase origin master`) and force-push, I am confident all goes green (not that it really matters, but would be a nice test).

test/xbmcgui.py
dagwieers: BTW You can turn this back into `type`. With the new xbmc framework update I have ignored the `invalid-name` checks for these modules. (This is in fact an issue with the original modules :disappointed:)

test/xbmc.py
dagwieers: Evaluating this again, I really think we should be using `special://masterprofile` instead by default. We probably don't want to make this profile-specific and keep it irrelative to the actual profile used.

dagwieers:
@horstle I did the rebase and pushed one commit with the changes I proposed here.

dagwieers:
I think we may want to reorganize the settings at some point.
Probably add a seperator, and rename 'Troubleshooting' into 'Expert'.

mediaminister:
If people with ARM devices can confirm this works in real life, I'm okay with merging this.

dagwieers:
I have tested this, but not with a device that hasn't sufficient disk space.

horstle:
Great to see this finally merged, thanks! I really hope I didn't break anything with the last changes.

dagwieers:
We still have time for testing before we release. It would be nice to test the following cases:
- Test with insufficient disk space
- Point to a new temporary path (potentially on other storage)
- Test with second storage that also does not have insufficient disk space
- Remove second storage with temporary path still pointing to it
- Test with non-existing temporary path

PR #27, opened 2018-06-01
Fixes a python "No such file or directory" error

mrfixit2001:
This output shows the error that this PR resolves. My system does not have "loop" OR "modinfo" so the check for it fails. This resolves the check error.
https://pastebin.com/4wqVCPcZ

### DIFF
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -227,7 +227,13 @@ class Helper(object):
 
     def _check_loop(self):
         """Check if loop module needs to be loaded into system."""
-        if not self._run_cmd(['modinfo', 'loop'])['success']:
+        loopcheck = False
+        try:
+            loopcheck = self._run_cmd(['modinfo', 'loop'])['success']
+        except:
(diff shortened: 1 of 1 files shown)


lib/inputstreamhelper.py
```diff
@@ -227,7 +227,13 @@ def _run_cmd(self, cmd, sudo=False):
 
     def _check_loop(self):
         """Check if loop module needs to be loaded into system."""
-        if not self._run_cmd(['modinfo', 'loop'])['success']:
+        loopcheck = False
+        try:
+            loopcheck = self._run_cmd(['modinfo', 'loop'])['success']
+        except:
```
dagwieers: We cannot accept bare exceptions, so we need to catch the exact Exception that is thrown here.

dagwieers: We cannot accept bare exceptions, so we need to catch the exact Exception that is thrown here.

mediaminister:
I fixed this in https://github.com/emilsvennesson/script.module.inputstreamhelper/pull/78

PR #28, opened 2018-06-01
ModProbe needs to run as shell on some systems

mrfixit2001:
I'm using buildroot, and modprobe doesn't run unless the check_output command is run using shell=True. This pull request adds an additional optional parmater to the run_cmd definition that tells it whether to run as shell, and then passes that parameter as true when calling modprobe for the checkloop method.

### DIFF
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -203,12 +203,12 @@ class Helper(object):
         self._log('Failed to calculate losetup offset.')
         return False
 
-    def _run_cmd(self, cmd, sudo=False):
+    def _run_cmd(self, cmd, sudo=False, useshell=False):
         """Run subprocess command and return if it succeeds as a bool."""
         if sudo and os.getuid() != 0 and self._cmd_exists('sudo'):
             cmd.insert(0, 'sudo')
         try:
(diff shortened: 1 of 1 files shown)


dagwieers:
Can you please rebase your branch and fix these conflicts?

lib/inputstreamhelper.py
```diff
@@ -203,12 +203,12 @@ def _chromeos_offset(self, bin_path):
         self._log('Failed to calculate losetup offset.')
         return False
 
-    def _run_cmd(self, cmd, sudo=False):
+    def _run_cmd(self, cmd, sudo=False, useshell=False):
```
dagwieers: I would like to see this name `shell=False` rather than `useshell'.

dagwieers:
@mediaminister I guess this one deserves a makeover as well ;-)

mediaminister:
Not sure about this pull request.

dagwieers:
It is likely that on some systems/environments **modprobe** requires a proper shell (possibly for environment variables or a proper PATH) that is only there when the command is ran from a shell. Since there is no impact on others by running it in a shell, I think it is better to trust the report.

mediaminister:
Okay, fixed in https://github.com/emilsvennesson/script.module.inputstreamhelper/pull/83

PR #29, opened 2018-06-01
Support 64-bit kernel with 32-bit userspace

mrfixit2001:
This updates the error to check that the arm64 kernel is running in a 64-bit userspace before it throws a warning.


dagwieers:
Can you please rebase your branch and fix the conflicts?

IMO This is fine to be merged.

mediaminister:
Fixed this in https://github.com/emilsvennesson/script.module.inputstreamhelper/pull/81

PR #30, opened 2018-06-24
check widevine support before all checks

vlmaksime:


dagwieers:
Can you please rebase your branch?

dagwieers:
Also, I am not sure if this change is even relevant, as the add-on should be using:
```python
    is_helper = inputstreamhelper.Helper(PROTOCOL, drm=DRM)
    if is_helper.check_inputstream():
```
in order to check if **inputstreamhelper** is needed. So the only potential improvement is making a quicker decision based on other criteria.

dagwieers:
This is weird, the resulting file looks fine, but the diff includes the proxy support?
Seems like a GitHub bug?

vlmaksime:
Now it looks correct

dagwieers:
How about https://github.com/emilsvennesson/script.module.inputstreamhelper/pull/30#issuecomment-522544889 ?

mediaminister:
`drm` is an optional argument, so this could also be `None`, so I think checking the value of `drm` makes sense.

dagwieers:
With the power vested in me, I'll merge :-D

Issue #32, opened 2018-08-02
Option to disable inputstreamhelper in settings + manual download cdm over GUI settings


JohnPlayerSpecial2018:
Please work in 2 settings over GUI (maybe in inputstream.adaptive addon settings). It improves usability and avoids workaround over ssh:

1) "Option to disable inputstreamhelper" -> Not everybody has enabled automatic addon updates and updating cdm on an older inputstream.adaptive can cause incompatibility. Would be more user-friendly to have this option also over GUI.

2) Manual download cdm -> If users reject cdm update once (because of staying on older/incompatible inputstream.adaptive version, temporal reasons, by mistake..), there is no other chance for updating over Kodi GUI as to wait until inputstreamhelper asks again (this can take long time..).

horstle:
I actually worked on something similar over here: [https://github.com/horstle/script.module.inputstreamhelper/tree/temp_disable](https://github.com/horstle/script.module.inputstreamhelper/tree/temp_disable)

I don't really like the code about manually running the helper and temporary disabling is currently not working if I remember correctly. Anyways making the "option to disable" visible is only a tiny change, I'll create a PR soon (probably today). After that it will be possible to change the setting by going to the kodi settings > system (change from standard to expert there) > addons > manage dependencies > Inputstream Helper > configure.

PR #34, opened 2018-08-15
Make option to disable module visible in GUI

horstle:
As stated in #32 with this change you can enable/disable inputstreamhelper via GUI at 
"kodi settings" > "system" (change from standard to expert there) > "addons" > "manage dependencies" > "Inputstream Helper" > "configure".

### DIFF
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -205,8 +205,12 @@ class Helper:
 
     def _helper_disabled(self):
         """Return if inputstreamhelper has been disabled in settings.xml."""
-        disabled = ADDON.getSettingBool('disabled')
-        if disabled:
+        disabled = ADDON.getSetting('disabled')
+        if not disabled:
+            ADDON.setSetting('disabled', 'false')  # create default entry
+            disabled = 'false'
+
+        if disabled == 'true':
             self._log('inputstreamhelper is disabled in settings.xml.')
             return True
 
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -205,19 +205,14 @@ class Helper:
 
     def _helper_disabled(self):
         """Return if inputstreamhelper has been disabled in settings.xml."""
-        disabled = ADDON.getSetting('disabled')
-        if not disabled:
-            ADDON.setSetting('disabled', 'false')  # create default entry
-            disabled = 'false'
-
-        if disabled == 'true':
+        disabled = ADDON.getSettingBool('disabled')
+        if disabled:
             self._log('inputstreamhelper is disabled in settings.xml.')
             return True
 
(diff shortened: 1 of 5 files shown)


horstle:
I guess breaking compatibility with Kodi versions before Leia is not wanted (yet)? If you want to I can revert the second commit, doesn't change the functionality anyway.

dagwieers:
I'd like to postpone this for the v0.5.0 release as this has an impact on how the plugin behaves.

One concern I have is that users do not have an friendly way to access this plugin's settings.
So this would then only be useful if (consumer) add-on provide a settings entry to access this one's.
And that will only be useful if the settings has a wider use.

PS Can you please rebase your PR?

horstle:
Done.

lib/inputstreamhelper.py
```diff
@@ -181,12 +181,8 @@ def _cmd_exists(cmd):
 
     def _helper_disabled(self):
         """Return if inputstreamhelper has been disabled in settings.xml."""
-        disabled = ADDON.getSetting('disabled')
-        if not disabled:
-            ADDON.setSetting('disabled', 'false')  # create default entry
-            disabled = 'false'
-
-        if disabled == 'true':
+        disabled = ADDON.getSettingBool('disabled')
```
vlmaksime: This method has been available since Kodi 18. It's broke backwards compatibility with Kodi 17 and later.

horstle: I asked whether I should revert this change before.
https://github.com/emilsvennesson/script.module.inputstreamhelper/pull/34#issuecomment-475909581

vlmaksime: > I'll have to create a new commit changing this back, if that's what you want.
Yes. I meant that. Thanks.

horstle: Reverted it.

horstle:
OK, the commits have been really fucked up by the rebase, but the end result is fine. I think there's no harm in merging this and building a full-fledged setting page based on it.

dagwieers:
I agree, this doesn't require any API calls. Let's get you out of this misery ;-)

horstle:
Thanks

dagwieers:
While reworking the interface for the future, I think we should redesign this interface.
Interfaces should list functionality, and functionality can be enabled or disabled.

The problem here is that "disabled" is actually a negative imposition. The value "true" should be enabled, and a value "false" should be disabled. And by default this functionality is enabled. This makes it a much better interface for users.

The problem with changing this is two-fold, either we support people that have used this manually and make it more complex and error-prone, or we don't, and those advanced users would need to reapply. IMHO I don't think we should be adding complexity, and expect that users do not have it disabled. (Especially with the recent Widevine CDM breakage, users have it enabled again ;-))

I will propose this in a subsequent PR.

PR #36, opened 2018-09-11
Creates backup of CDM before installing new version and...

horstle:
...adds function to revert (is_helper.revert_cdm()). Maybe could be called by inputstream if there is an error?

### DIFF
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -440,6 +440,16 @@ class Helper(object):
 
         return devices
 
+    def _backup_cdm(self):
+        """Loops through local cdm folder and creates backup."""
+        for cdm_file in os.listdir(self._addon_cdm_path()):
+            if cdm_file.endswith(config.CDM_EXTENSIONS):
+                self._log('Backing up file: {0}'.format(cdm_file))
+                cdm_file_current = os.path.join(self._addon_cdm_path(), cdm_file)
+                shutil.copyfile(cdm_file_current, cdm_file_current + '.bak')
+
+        return True
+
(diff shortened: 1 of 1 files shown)


dagwieers:
I would like to take this one step further:
- Keep a number (say 5) backups of Widevine CDM versions
- Offer the option to install a Widevine CDM manually (so you can revert to any version)
- Show a selection of versions (based on versions.txt and what is on disk)

horstle:
I could probably implement all that after #87 is done. Just not yet sure how to read the available versions in the settings page.

dagwieers:
@horstle At this time I wouldn't spend too much time on it, we first need to get the basics right.

For me the absolute minimum is:
- Proper Widevine CDM version reporting (on ARM)
  - This requires we store both the ChromeOS image version as well as Widevine CDM version
- Interface to disable **inputstreamhelper**
- Interface to remove the currently used Widevine CDM library
- Interface to (re)install the Widevine CDM library

IMO we probably want to include the above as part of the upcoming release, given that it requires a minimum of effort and already covers 99% of the user's needs.

The need for backups and rolling back is only useful when we (ever) get a broken Widevine CDM. And we should be looking at #69 to make this case more user-friendly first. Otherwise you will be needing to rebase and resolve conflicts all the time :-/

dagwieers:
I opened a new feature request for this #165 with some more details.

dagwieers:
This relates to #165. I don't mind keeping this one open in case @horstle wants to reuse this branch for this work ;-)

horstle:
Closing this in favor of #193.

PR #42, opened 2019-02-01
Support for Unicode paths in Windows

WallyCZ:
Currently plugin crashes when there is some Unicode character in users profile path.
`WindowsError: [Error 3] The system cannot find the path specified: 'C:\\Users\\V\xc3\xa1clav\\AppData\\Roaming\\Kodi\\cdm/*.*'`

This PR should fix it - convert this path to utf-8 at first, because some functions in python 2 expecting path in utf-8 (or something like that :) )


WallyCZ:
Probably related to this: https://forum.kodi.tv/showthread.php?tid=125492

emilsvennesson:
Thank you!

PR #51, opened 2019-03-22
Add proxy support (on-the-fly)

### LINKED ISSUE #50: [Enhancement] Add proxy support
While testing our own addon's proxy support I discovered that **inputstreamhelper** does not adhere to Kodi's proxy configuration, which breaks playing DRM streams.

It is not that hard to make addon's using Requests to use Kodi's proxy settings. You can find our implementation at: https://github.com/pietje666/plugin.video.vrt.nu/pull/87

The testing methodology I used was to disable the default gateway (or in LibreELEC, configure a non-existing gateway) and then test your local proxy and see if everything works as expected.

dagwieers:
This PR adds proxy support to the requests call using the global Kodi
proxy settings.

This fixes #50

### DIFF
--- a/addon.xml
+++ b/addon.xml
@@ -3,6 +3,7 @@
     <requires>
         <import addon="xbmc.python" version="2.25.0"/>
         <import addon="script.module.kodi-six" version="0.0.4"/>
+        <import addon="script.module.pysocks" version="1.6.8" optional="true"/>
     </requires>
     <extension library="lib" point="xbmc.python.module"/>
     <extension point="xbmc.addon.metadata">
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -23,6 +23,12 @@ import config
 
 from kodi_six import xbmc, xbmcaddon, xbmcgui, xbmcvfs
 
+try:
+    import socks  # pylint: disable=unused-import
+    HAS_SOCKS = True
+except ImportError:
+    HAS_SOCKS = False
+
 ADDON = xbmcaddon.Addon('script.module.inputstreamhelper')
 ADDON_PROFILE = xbmc.translatePath(ADDON.getAddonInfo('profile'))
 LANGUAGE = ADDON.getLocalizedString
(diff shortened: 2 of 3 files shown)


dagwieers:
Code is taken from my contribution (pietje666/plugin.video.vrt.nu#87) to another Kodi addon.

dagwieers:
I tested this code successfully, downloading the widevine 2GB over a local-network proxy.

dagwieers:
I rebased and force-pushed.

dagwieers:
This needs to be redone with urllib/urllib2 support after #71

basrieter:
Won't this work out of the box? If you configure Kodi to use proxies, the Python stuff also uses that proxy.

dagwieers:
No it doesn't.

basrieter:
I doubt that. As I always use Fiddler to debug and I see the requests from the InputStream helper coming in.

dagwieers:
I can confirm you this is not the case. At least not for the VRT NU plugin, so strictly hypothetical it may be because Retrospect is modifying the layer that inputstreamhelper is depending on?

basrieter:
For 100%: If I configure Kodi to use a proxy, that proxy is used by Python as wel. At least on Windows.

dagwieers:
@basrieter It is not true, not even on Windows. I just verified on Windows 10 with Kodi 18.3 using both the VRT NU and Retrospect add-ons. They don't use a proxy if you download the Widevine CDM. I have a dedicated proxy for testing and none of the downloads show up when I configure a proxy at the Kodi level.

For the same reason we have to modify our add-ons so that they take the Kodi proxy configuration into account for making HTTP requests, we have to modify inputstreamhelper. There is no magic involved that somehow takes the proxy configuration and injects it into requests or urllib/urllib2.

The only possibility is that Kodi would set `http_proxy` environment variables for python, but it does not seem to be doing that, and rightly so, as it wouldn't just work for everything (e.g. SOCKS, etc.)

dagwieers:
So, after testing with @basrieter, Kodi does set the http_proxy variable for proxy type 0, but it requires a restart/reboot to work. That is why in my testing it doesn't work for inputstreamhelper, but it does work for the add-ons (that implement proxy support on-the-fly).

The use of http_proxy does not work for other proxy types, like SOCKS.

dagwieers:
Validated with current VRT NU and VTM GO code, and tested with a local proxy.

PR #53, opened 2019-03-28
auto install inputstream.adaptive when not installed

mediaminister:
When inputstream.adaptive is not installed, install it.
Because of a bug upstream, for now you have to restart Kodi to get the binary addon inputstream.adaptive working.
But this Kodi bug is already fixed in de nightly builds: https://github.com/xbmc/xbmc/pull/15780

### DIFF
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -752,14 +752,27 @@ class Helper(object):
 
         return True
 
+    def _install_inputstream(self):
+        """Install inputstream addon."""
+        try:  # See if there's an installed repo that has it
+            xbmc.executebuiltin('InstallAddon({})'.format(self.inputstream_addon), True)
+            addon = xbmcaddon.Addon('{}'.format(self.inputstream_addon))
+            self._log('inputstream addon installed from repo')
+            return True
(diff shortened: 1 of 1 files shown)


lib/inputstreamhelper.py
```diff
@@ -752,14 +752,27 @@ def _check_drm(self):
 
         return True
 
+    def _install_inputstream(self):
+        """Install inputstream.adaptive."""
+        try:  # See if there's an installed repo that has it
+            xbmc.executebuiltin('InstallAddon(inputstream.adaptive)', True)
```
emilsvennesson: Please change the code so it works for other inputstream add-ons as well.
```python
xbmc.executebuiltin('InstallAddon({})'.format(self.inputstream_addon), True)
addon = xbmcaddon.Addon('{}'.format(self.inputstream_addon))
```

mediaminister:
@emilsvennesson  On April 22, Kodi 18.2 was released which enables realtime installing binary add-ons like InputStream Adaptive without restarting Kodi.
Can you please make a new release?

emilsvennesson:
Will do ASAP.

PR #61, opened 2019-08-14
update widevine when new inputstreamhelper version first runs

mediaminister:
This automatically updates widevine when inputstreamhelper is updated and running for the first time

### DIFF
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -553,10 +553,27 @@ class Helper(object):
 
         return False
 
+    def _first_run(self):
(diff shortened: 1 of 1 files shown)


dagwieers:
@mediaminister Thanks !

There's one suggestion. I would use `version` for the addon version, and would move to `widevine_version` (non-ARM) and `chromeos_version` (ARM) to track the actualy Widevine CDM.

mediaminister:
That's a possibility for the future. I followed the most simple approach for now. Getting a functional inputstreamhelper version released is more important now.

Issue #63, opened 2019-08-14
Add unit and integration tests


dagwieers:
Because this addon is important for a lot of Kodi functionality, we need to ensure we can rely on it to work, and keeps on working. For this we definitely need unit tests and integration tests, and run them regularly so we can prevent and report known issues before they cause issues for users.

Also, when we have unit tests and integration tests, it makes it possible to test the performance and overhead of the plugin as well.

dagwieers:
@mediaminister Not sure how we want to do this. Do we use Mock for this, or do we reuse our testing framework from VRT NU?

mediaminister:
I think we can reuse a subset of the VRT NU testing framework.

PR #65, opened 2019-08-15
Add sanity tests

### LINKED ISSUE #64: Remove requirement on urllib3 and requests
Since **urllib3** and **requests** is very slow to be imported on Raspberry Pi, it would be better to remove it from the source code and implement **urllib**/**urllib2** support. Especially since the code is pretty simple and doesn't need any special functionality from requests.

dagwieers:
This PR includes:
- Add sanity testing, including Travis
- Fix all pylint and PEP8 issues
- Add pip requirements.txt
- Add a Makefile for creating a working ZIP package

This relates to #63

### DIFF
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -786,7 +786,7 @@ class Helper:
             xbmc.executebuiltin('InstallAddon({})'.format(self.inputstream_addon), True)
 
             # Check if InputStream add-on exists!
-            addon = xbmcaddon.Addon('{}'.format(self.inputstream_addon))  # pylint: disable=unused-variable
+            xbmcaddon.Addon('{}'.format(self.inputstream_addon))
 
             self._log('inputstream addon installed from repo')
             return True
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -177,7 +177,8 @@ class Helper:
             self._log('inputstreamhelper is disabled in settings.xml.')
             return True
 
-        self._log('inputstreamhelper is enabled. You can disable inputstreamhelper by setting \"disabled\" to \"true\" in settings.xml (Note: only recommended for developers knowing what they\'re doing!)')
+        self._log('inputstreamhelper is enabled. You can disable inputstreamhelper by setting \"disabled\" to \"true\" in settings.xml \
+        (Note: only recommended for developers knowing what they\'re doing!)')
         return False
 
     def _inputstream_version(self):
@@ -330,8 +331,8 @@ class Helper:
                     progress_dialog.update(percent)
             progress_dialog.close()
             return True
-        else:
-            return req.text
+
+        return req.text
 
     def _has_inputstream(self):
         """Checks if selected InputStream add-on is installed."""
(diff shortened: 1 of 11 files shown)


dagwieers:
It would be nice if Travis integration could be enabled for the sanity testing.

dagwieers:
I am sorry for mixing 2 things up in a single PR, but I started with removing Requests with urlib/urllib2 for performance reasons (removing 800ms hit for loading urllib3 on RPi).

And only when I had that done, I looked at testing this code easily which required some changes. An so I added PEP8 and pylint to this as well.

mediaminister:
I checked your changes and saw you broke `_install_inputstream` function. Because `xbmc.executebuiltin` doesn't return if it succeeded or failed we have to check afterwards if the inputstream addon exists. I added this fix to this pull request already.

I will do a couple of more tests today and report back.

emilsvennesson:
Please split the changes into separate commits before merging as well.

lib/inputstreamhelper.py
```diff
@@ -801,9 +801,13 @@ def _check_drm(self):
 
     def _install_inputstream(self):
         """Install inputstream addon."""
```
dagwieers: Thanks! This must have been a merge issue, as this was properly tested.

Travis would have caught this too 😉

dagwieers:
@emilsvennesson How do you want them to be split?

lib/inputstreamhelper.py
mediaminister: No, it was not a merge issue. I didn't see the error in the sanity tests, I did a manual test with a LibreELEC virtual machine. This problem occurs **after** `xbmc.exeutebuiltin` fails to install the inputstream add-on. Later on in the script inputstreamhelper is doing a version check on an unexisting add-on and that causes a crash.

dagwieers: On my cellphone it appeared as if the try-statement was missing in my code.
That's why I thought this was basically a syntax error in the first place.

dagwieers:
@emilsvennesson I don't mind separating the Requests-change from the sanity checks/fixes. (As it should have been), but separating the different commits make no sense as these are gradual fixes during development. Each change does not represent a new entity, so using the reflog to turn them in commits is a waste of everyone's time.

The amend-commits during development (in most cases) are there for testing, to ensure a new commit hash is used in the ZIP file so I can actually test individual changes, without having a new commit for every small change or typo.

emilsvennesson:
@dagwieers Yeah, it was the requests change I'd like to see as a separate commit. I'm with you on the rest.

dagwieers:
I think I managed to split both. It was dirty :-(
And now it needs to be retested, twice.

mediaminister:
Added a couple of fixes. Looks good to me now.

dagwieers:
I am going to merge this as it is holding everything back at the moment.

PS We could set rules on this repository that e.g. requires PRs to be reviewed or CI turns green before they can be merged etc.

basrieter:
Seems reasonable

mediaminister:
I'm okay with setting rules, but I guess that's something the repo owner has to do.

dagwieers:
@mediaminister Yes, same for Travis. Only the owner can do this :-/

PR #68, opened 2019-08-15
convert Dutch strings.po to UTF-8

mediaminister:
The file was previously encoded in Western ISO
UTF-8 is a requirement for official Kodi repo.


dagwieers:
I noticed the same thing. Was on my list 😉

PR #71, opened 2019-08-16
Replace requests and urllib3 with urllib/urllib2

### LINKED ISSUE #64: Remove requirement on urllib3 and requests

dagwieers:
This PR includes:
- Replace requests library with urllib/urllib2
- Increase download chunk size by factor 32 to speed up transfer

This fixes #64

### DIFF
--- a/addon.xml
+++ b/addon.xml
@@ -3,7 +3,6 @@
     <requires>
         <import addon="xbmc.python" version="2.25.0"/>
         <import addon="script.module.kodi-six" version="0.0.4"/>
-        <import addon="script.module.requests" version="2.9.1"/>
     </requires>
     <extension library="lib" point="xbmc.python.module"/>
     <extension point="xbmc.addon.metadata">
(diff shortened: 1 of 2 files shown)


mediaminister:
There was a problem with installing Widevine CDM in Kodi python3 because bytes and unicode were mixed while generating the download url. We should always decode bytes to unicode when getting text input.
Fixed this and found no other problems. For me, this can be merged.

dagwieers:
@emilsvennesson How do you see this going forward? What are the rules for merging PRs?
Can you enable Travis-support so we can test contributions automatically?

emilsvennesson:
I have enabled Travis CI. Please tell me if there's anything else you'd like me to do.

Personally, as long as it's been properly tested I have no problems with you merging PR:s once you feel like it's ready.

Issue #35, opened 2018-09-05
Odroid C2 Unable to load shared library


rgainza:
```log
19:33:54.179 T:3975140208   ERROR: AddOnLog: InputStream Adaptive: Unable to load widevine shared library (/storage/.kodi/cdm/libwidevinecdm.so)
```

Your script downloads the widevine module. No problem there, but for some reason I can't get that module to work. The macleod script worked for it without any problems so I thought it was the same structure as the RPi. 

Which recovery OS image is the best for the Odroid C2?

deorder:
Temp fix for the ARM version. 7zip or another archive tool that supports filesystem images required.

Go to: https://cros-updates-serving.appspot.com/#speedy

Click `67` on the line with the codename `Speedy`. It will download an older version of Chrome OS.
Direct link: https://dl.google.com/dl/edgedl/chromeos/recovery/chromeos_10575.58.0_daisy-spring_recovery_stable-channel_spring-mp-v3.bin.zip

Open the archive, open the `.bin` file and then open the `ROOT-A.img` file. You will several directories.

Go to `\opt\google\chrome\`. In there you will find the file `libwidevinecdm.so`.

Extract the file `libwidevinecdm.so` and copy it to your Kodi home directory in `~/.kodi/userdata/addon_data/script.module.inputstreamhelper/cdm` or similar (instead of overwriting you can backup the already existing one).

Restart Kodi

rgainza:
Thank you!

rspierenburg:
I download the file, extract the zip to a .bin, but when I try to extract the .bin file it tells me there is an error in the file.  I've downloaded both from the direct link and the actual page.  Is there anywhere else I can get the libwidevinecdm.so file?

rgainza:
I don't think it's allowed to share the .so file openly according to the license (copying not allowed)? 

Just use the SSH and follow the instructions. 
https://help.vivaldi.com/article/raspberry-pi/#drm-and-flash

You could create a bash script:

#!/bin/sh
CHROMEOSR="https://dl.google.com/dl/edgedl/chromeos/recovery/chromeos_10575.58.0_daisy-spring_recovery_stable-channel_spring-mp-v3.bin.zip"
wget $CHROMEOSR
unzip $(basename $CHROMEOSR)
LOOPD=$(losetup -f)
sudo losetup -P $LOOPD $(basename $CHROMEOSR .zip)
mkdir chromeos
[...]

PR #75, opened 2019-08-19
Remove legacy support

### LINKED ISSUE #66: Remove legacy code
The codebase seems to have some support for legacy Kodi, but the Widevine CDM (version **1.4.8.903**) used for legacy Kodi does not seem to be supported any longer. So maybe it is wise to remove the legacy code from the source ?

See: https://www.widevine.com/news

dagwieers:
Since the legacy widevine version is no longer supported, and the newer
widevine versions are only supported by Kodi v18, there is no need to
keep legacy support around.

This fixes #66 
This relates to #35


mediaminister:
No problems found, ready to merge.

PR #81, opened 2019-08-19
Support 64-bit kernel with 32-bit userspace

mediaminister:
This updates the error to check that the arm64 kernel is running in a 64-bit userspace before it throws a warning.

### DIFF
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -12,6 +12,7 @@ import shutil
 import re
 from distutils.version import LooseVersion
 from datetime import datetime, timedelta
+import struct
 
 try:  # Python 3
     from urllib.error import HTTPError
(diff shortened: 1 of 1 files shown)


dagwieers:
I'd rather get the original fixes in to show some respect to people that have been waiting for more than a year to get it merged ;-)

dagwieers:
Can you please change the author of your commit (by amending your commit with the original author)? So that the original author gets the credit.

dagwieers:
Thanks!

mediaminister:
I never changed the author. Seems like a github bug?

dagwieers:
No no, this is fine. In the past it would only show the author, but that is a security concern (you could fake someone else authored something). Now it shows both, but that is fine.

> mrfixit2001 authored and mediaminister committed 20 minutes ago

Issue #62, opened 2019-08-14
Make Widevine ARM selection more reliable


dagwieers:
So #59/#60 happened on ARM because the selected hardware device (codename SPRING) is no longer (or temporarily not) getting newer recovery images, and as such does not ship a newer, supported Widevine CDM library. (The only ARM device in the list BTW)

Rather than having a single hardware device listed in our *config.py*, we could instead list all known ARM devices, and have an algorithm select the smallest ARM image that has the latest version available.

I went over the long list of ChromeOS devices and verified the hardware specs to see which ones in that list actually ship with an ARM processor. This is the list:

```
# 1101940875  12105.100.0  BOB       // ASUS Chromebook Flip C101PA
# 1219529228  12105.100.0  WHITETIP  // CTL Chromebook J41 / J41T
#  602050152  12105.100.0  SKATE     // HP Chromebook 11 G2
#  589868247  11316.165.0  SPRING    // HP Chromebook 11 G1
#  602667331  12105.100.0  SNOW      // Samsung Chromebook - XE303
# 1102818796  12105.100.0  ELM       // Acer Chromebook R13 (CB5-312T)
# 1102001394  12105.100.0  HANA      // *** (Lenovo, Poin2)
#  613700233  12105.100.0  BIG       // Acer Chromebook 13 (CB5-311, C810)
#  614552734  12105.100.0  BLAZE     // HP Chromebook 14 G3
# 1174046692  12105.100.0  RELM      // *** (Acer, CTL, Edxis, HP, Mecer, Positivo)
# 1110369689  12105.100.0  DUMO      // ASUS Chromebook Tablet CT100
# 1110369689  12105.100.0  SCARLET   // Acer Chromebook Tab 10
# 1087032125  12105.100.0  FIEVEL    // AOpen Chromebox Mini
#  600449940  12105.100.0  JAQ       // *** (Haier, Lava, Medion, True)
# 1087717823  12105.100.0  JERRY     // *** (CTL, EduGear, Epik, HiSense, Mecer, NComputing, Poin2, Positivo, VideoNet)
#  600853103  12105.100.0  MICKEY    // ASUS Chromebit CS10
# 1087997726  12105.100.0  MIGHTY    // *** (EduGear, Haier, Lumos, MEDION, Nexian, PCMerge, Prowise, Sector 5, Viglen)
# 1089562352  12105.100.0  MINNIE    // ASUS Chromebook Flip C100PA
[...]

Issue #72, opened 2019-08-17
Access the zipped image directly to extract CDM


dagwieers:
There are four major issues with inputstreamhelper on ARM:
- The need for root privileges in order to mount the disk image
- The required tools and kernel capability for performing the process
- The need for 2GB of free disk space
- The time to uncompress a 600MB image on a RPi and write 2GB to an SD card

We can avoid all of this, by using FUSE to mount as a user, and using zipfs to mount the zipped image directly without the need to extract the complete image.

There is a fusepy project to create python-based fuse modules, and there is a zipfs fuse module written in python as well. Currently this enables us to access the binary without the need to uncompress the ZIP file, but it does not allow us to mount the filesystem from without this mounted ZIP file. So what we need to do is create our own extzipfs that does both in a single mount. Or alternatively, instead of mounting it, use the fuse interface to get that file directly from the zipped ext2 image.

This would save us a great deal of time as well, as we wouldn't have to write to the SD card as heavily as we do now.

dagwieers:
In fact FAT.py from FATtools is very close to what we need already, we simply have to wrap the file accessing functions around zipfile. So even FUSE is not needed!

dagwieers:
So apparently the ZIP file is a disk image with a GPT MBR, and various partitions, including an ext2 partition with a filesystem that stores the libwidevinecdm.so.

So our tool needs to be able to:
- access the ZIP file as a normal file (block-device)
- parse the GPT MBR to find the correct partition offset
- access the ext2 filesystem from that offset
- walk the tree to find a file by name

dagwieers:
I have succeeded to mount a partition image using pyext2 and extract our file (with an exact path).
This does not yet:
- traverse the tree to find a file by base name
- read from a specific offset of an image
- work on a zipped image

The two last features would require changes to pyext2. Even if we cannot work directly on a zipped image, being able to extract from the image would already avoid needing to mount as root.

If we go down this path, we will be actively working with the pyext2 project and would need to release a pyext2 and a script.module.pyext2 plugin.

dagwieers:
Uncompressing the zipped image while downloading (to save on disk space) is not an option. with python's ZipFile as it requires the ZipFile to be downloaded completely to memory first. There is no stream interface.

Having random access to the compressed Zip image will be hard as python's ZipFile has no random access interface either. So it would require a new library with an mmap() interface.

dagwieers:
[...]

PR #82, opened 2019-08-19
Add proper progress dialog for extraction

### LINKED ISSUE #70: Provide a progress bar for the ARM extraction/copy phase
Currently, one of the issues users have is that the second phase of the ARM Widevine CDM installation process does not show any indication it is still performing actions. (And people tend no to read or understand pop-up windows).

So I think it is better that we replace the warning OK-box with a progress-dialog that for each step shows progress and ensure that the user cannot do anything (except interrupt/end the process), rather than run another re-install while the previous has not ended yet.

dagwieers:
This PR includes:
- Replace the defunct `DialogBusy()` with a `DialogProgress()` window
- Remove unneeded OK windows
- Add short English translation messages to make code more readable
- Includes small fixes for #62 

This fixes #70

### DIFF
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -582,6 +582,10 @@ class Helper:
         root_cmds = ['mount', 'umount', 'losetup', 'modprobe']
         devices = self._chromeos_config()
         arm_device = self._select_best_chromeos_image(devices)
+        if arm_device is None:
+            self._log('We could not find an ARM device in recovery.conf')
+            xbmcgui.Dialog().ok(LANGUAGE(30004), LANGUAGE(30005))
+            return ''
         required_diskspace = int(arm_device['filesize']) + int(arm_device['zipfilesize'])
         if xbmcgui.Dialog().yesno(LANGUAGE(30001),  # Due to distributing issues, this takes a long time
                                   LANGUAGE(30006).format(self._sizeof_fmt(required_diskspace))) and self._widevine_eula():
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -551,8 +551,8 @@ class Helper:
 
         downloaded = self._http_download()
         if downloaded:
-            busy_dialog = xbmcgui.DialogProgress()
-            busy_dialog.create()
+            progress_dialog = xbmcgui.DialogProgress()
+            progress_dialog.create()
             self._unzip(self._addon_cdm_path())
 
             if not self._widevine_eula():
@@ -569,10 +569,10 @@ class Helper:
                 wv_check = self._check_widevine()
                 if wv_check:
                     xbmcgui.Dialog().notification(LANGUAGE(30037), LANGUAGE(30003))  # Widevine successfully installed
-                    busy_dialog.close()
+                progress_dialog.close()
                 return wv_check
 
-            busy_dialog.close()
+            progress_dialog.close()
             xbmcgui.Dialog().ok(LANGUAGE(30004), LANGUAGE(30005))  # An error occurred
 
         return False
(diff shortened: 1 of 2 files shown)


dagwieers:
I have one issue where it (sometimes) crashes Kodi after successfully having installed Widevine CDM.

dagwieers:
The progress bar does not move during uncompressing the image. This however could be possible, since we also get the uncompressed image size from the recovery images conf file. But it would require either that we unzip it from Python (probably a good thing) or we monitor the uncompressing of the file. (I'd rather implement #72)

dagwieers:
This is working quite well to the point that I would like to merged this as-is.
Please verify the text strings and interface to the user so we are sure this is what we want to report.
Only then can we start looking at translating these new strings.

resources/language/resource.language.en_gb/strings.po
```diff
@@ -180,3 +180,35 @@ msgstr ""
 msgctxt "#30042"
 msgid "Using a SOCKS proxy requires the PySocks library (script.module.pysocks) installed."
 msgstr "Using a SOCKS proxy requires the PySocks library (script.module.pysocks) installed."
+
+msgctxt "#30043"
+msgid "Extracting Widevine CDM"
+msgstr ""
+
+msgctxt "#30044"
+msgid "Prepare downloaded image..."
```
mediaminister: For` xbmcgui.DialogProgress()` messages I think it's better to use the present participle of the verb: preparing

dagwieers: Agreed.

dagwieers:
Let me know when you have tested this and you approve.

As this is not the final phase, the current dialogs are quite confusing/frightening for non-technical people, to the point that they do not install Widevine. ("I don't know what this is?", "Is this safe?")

One of the problems is that the decisions buttons default to "Cancel", which enforces the feeling that this may be something bad, or something you need to understand thoroughly before continuing. (Rightly so ;-)) So we need to make the text simple, friendly and informative.

lib/inputstreamhelper.py
```diff
@@ -612,38 +613,39 @@ def _install_widevine_arm(self):
...
                         with open(self._widevine_config_path(), 'w') as config_file:
                             config_file.write(json.dumps(devices, indent=4))
-                        xbmcgui.Dialog().notification(LANGUAGE(30037), LANGUAGE(30003))
-                        busy_dialog.close()
                         wv_check = self._check_widevine()
                         if wv_check:
-                            xbmcgui.Dialog().notification(LANGUAGE(30037), LANGUAGE(30003))
-                            busy_dialog.close()
+                            progress_dialog.update(99, line1=LANGUAGE(30003))  # Widevine installation was successful
+                        xbmcgui.Dialog().notification(LANGUAGE(30037), LANGUAGE(30003))  # Widevine installation was successful
```
dagwieers: I am not so convinced about this part.

The original code send a *success* notification before doing `self._check_widevine()` and after the check (when it was successful). So basically always, and sometimes twice. I guess that was wrong.

But I don't know what we should do if `self._checkwidevine()` would return False...

dagwieers: It originates from this: https://github.com/emilsvennesson/script.module.inputstreamhelper/commit/c50904cc9c85dfd413df908c0ba6c4f9643720f5

lib/inputstreamhelper.py
```diff
@@ -550,8 +551,8 @@ def _install_widevine_x86(self):
 
         downloaded = self._http_download()
         if downloaded:
-            busy_dialog = xbmcgui.DialogBusy()
-            busy_dialog.create()
+            progress_dialog = xbmcgui.DialogProgress()
+            progress_dialog.create()
```
mediaminister: `heading` is a required argument for  `xbmcgui.DialogProgress().create()`

```
2019-08-20 20:47:55.981 T:140252882970368   ERROR: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                             - NOTE: IGNORING THIS CAN LEAD TO MEMORY LEAKS!
                                            Error Type: <type 'exceptions.TypeError'>
                                            Error Contents: Required argument 'heading' (pos 1) not found
                                            Traceback (most recent call last):
                                              File "/storage/.kodi/addons/plugin.video.vrt.nu/resources/lib/addon_plugin.py", line 8, in <module>
                                                run(sys.argv)
                                              File "/storage/.kodi/addons/plugin.video.vrt.nu/resources/lib/addon.py", line 254, in run
                                                plugin.run(argv)
                                              File "/storage/.kodi/addons/script.module.routing/lib/routing.py", line 115, in run
                                                self._dispatch(path)
                                              File "/storage/.kodi/addons/script.module.routing/lib/routing.py", line 126, in _dispatch
                                                view_func(**kwargs)
                                              File "/storage/.kodi/addons/plugin.video.vrt.nu/resources/lib/addon.py", line 226, in play_id
                                                VRTPlayer(kodi).play(dict(video_id=video_id, publication_id=publication_id))
                                              File "/storage/.kodi/addons/plugin.video.vrt.nu/resources/lib/vrtplayer.py", line 287, in play
                                                self._kodi.play(stream, video.get('listitem'))
                                              File "/storage/.kodi/addons/plugin.video.vrt.nu/resources/lib/kodiwrapper.py", line 256, in play
                                                if is_helper.check_inputstream():
                                              File "/storage/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper.py", line 917, in check_inputstream
                                                return self._check_drm()
                                              File "/storage/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper.py", line 871, in _check_drm
                                                return self._install_widevine()
                                              File "/storage/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper.py", line 660, in _install_widevine
                                                return self._install_widevine_x86()
                                              File "/storage/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper.py", line 555, in _install_widevine_x86
                                                progress_dialog.create()
                                              File "/storage/.kodi/addons/script.module.kodi-six/libs/kodi_six/utils.py", line 62, in wrapper
                                                return py2_decode(func(*mod_args, **mod_kwargs))
                                            TypeError: Required argument 'heading' (pos 1) not found
                                            -->End of Python script error report<--
```
https://romanvm.github.io/Kodistubs/_autosummary/xbmcgui.html#xbmcgui.DialogProgress

dagwieers: Thanks, fixed.

dagwieers:
I am going to go forward with merging this one as it is hampering my testing efforts.
If any issue arise from this, we'll fix them.

PR #83, opened 2019-08-19
ModProbe needs to run as shell on some systems

mediaminister:
I'm using buildroot, and modprobe doesn't run unless the check_output command is run using shell=True. This pull request adds an additional optional parmater to the run_cmd definition that tells it whether to run as shell, and then passes that parameter as true when calling modprobe for the checkloop method.


dagwieers:
Thanks again. We are going fast :-)

PR #84, opened 2019-08-20
Select the best ARM image available

### LINKED ISSUE #62: Make Widevine ARM selection more reliable

dagwieers:
Rather than pre-selecting a specific device (which may get unsupported
or stale suddenly) it is better that we select an ARM device image that
fits our purpose: the latest version that is the smallest in size

This implements this better selection, and instead of `SKATE` we now use `BIG`.
Since the `SKATE` image is now older than `BIG`. `SKATE` is now stale.

This fixes #62

### DIFF
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -449,6 +449,41 @@ class Helper:
 
         return True
 
+    def _select_best_chromeos_image(self, devices):
+        best = None
+        for device in devices:
+            # Select ARM hardware only
+            for arm_hwid in config.CHROMEOS_RECOVERY_ARM_HWIDS:
+                if arm_hwid in device['hwidmatch']:
+                    hwid = arm_hwid
+                    break  # We found and ARM device
+            else:
+                continue  # Not ARM, skip this device
+
+            device['hwid'] = hwid
+
+            # Select the first ARM device
+            if best is None:
+                best = device
+                continue  # Go to the next device
+
+            # Skip identical hwid
+            if hwid == best['hwid']:
+                continue
+
+            # Select the newest version
+            if LooseVersion(device['version']) > LooseVersion(best['version']):
+                self._log('{device[hwid]} ({device[name]}) is newer than {best[hwid]} ({best[name]})'.format(device=device, best=best))  # pylint: disable=invalid-format-index
+                best = device
+
+            # Select the smallest image (disk space requirement)
+            elif LooseVersion(device['version']) == LooseVersion(best['version']):
+                if int(device['filesize']) + int(device['zipfilesize']) < int(best['filesize']) + int(best['zipfilesize']):
+                    self._log('{device[hwid]} ({device[name]}) is smaller than {best[hwid]} ({best[name]})'.format(device=device, best=best))  # pylint: disable=invalid-format-index
+                    best = device
+
+        return best
+
     def _latest_widevine_version(self, eula=False):
         """Returns the latest available version of Widevine CDM/Chrome OS."""
         if eula:
@@ -462,21 +497,26 @@ class Helper:
             versions = self._http_get().decode()
             return versions.split()[-1]
 
-        return [x for x in self._chromeos_config() if config.CHROMEOS_ARM_HWID in x['hwidmatch']][0]['version']
+        devices = self._chromeos_config()
+        arm_device = self._select_best_chromeos_image(devices)
+        return arm_device['version']
 
     def _chromeos_config(self):
         """Parses the Chrome OS recovery configuration and put it in a dictionary."""
-        devices = []
         self._url = config.CHROMEOS_RECOVERY_URL
         conf = [x for x in self._http_get().decode().split('\n\n') if 'hwidmatch=' in x]
+
+        devices = []
         for device in conf:
-            device_dict = {}
+            device_dict = dict()
             for device_info in device.splitlines():
-                key_value = device_info.split('=')
-                key = key_value[0]
-                if len(key_value) > 1:  # some keys have empty values
-                    value = key_value[1]
+                if not device_info:
+                    continue
+                try:
+                    key, value = device_info.split('=')
                     device_dict[key] = value
+                except ValueError:
+                    continue
             devices.append(device_dict)
 
         return devices
@@ -519,10 +559,10 @@ class Helper:
     def _install_widevine_arm(self):
         """Installs Widevine CDM on ARM-based architectures."""
         root_cmds = ['mount', 'umount', 'losetup', 'modprobe']
-        cos_config = self._chromeos_config()
-        device = [x for x in cos_config if config.CHROMEOS_ARM_HWID in x['hwidmatch']][0]
-        required_diskspace = int(device['filesize']) + int(device['zipfilesize'])
-        if xbmcgui.Dialog().yesno(LANGUAGE(30001),
+        devices = self._chromeos_config()
+        arm_device = self._select_best_chromeos_image(devices)
+        required_diskspace = int(arm_device['filesize']) + int(arm_device['zipfilesize'])
+        if xbmcgui.Dialog().yesno(LANGUAGE(30001),  # Due to distributing issues, this takes a long time
                                   LANGUAGE(30006).format(self._sizeof_fmt(required_diskspace))) and self._widevine_eula():
             if self._os() != 'Linux':
                 xbmcgui.Dialog().ok(LANGUAGE(30004), LANGUAGE(30019).format(self._os()))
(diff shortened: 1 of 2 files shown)


dagwieers:
We may have to update the list of ARM devices from time to time to ensure we are not running out of acceptable devices over time. (although older images tend to be smaller...)

lib/inputstreamhelper.py
```diff
@@ -449,6 +449,41 @@ def _supports_widevine(self):
...
         return True
 
+    def _select_best_chromeos_image(self, devices):
+        best = None
+        for device in devices:
+            # Select ARM hardware only
+            for arm_hwid in config.CHROMEOS_RECOVERY_ARM_HWIDS:
+                if arm_hwid in device['hwidmatch']:
+                    hwid = arm_hwid
+                    break  # We found and ARM device
```
mediaminister: Typo in comment?

lib/inputstreamhelper.py
```diff
@@ -462,21 +497,26 @@ def _latest_widevine_version(self, eula=False):
             versions = self._http_get().decode()
             return versions.split()[-1]
 
-        return [x for x in self._chromeos_config() if config.CHROMEOS_ARM_HWID in x['hwidmatch']][0]['version']
+        devices = self._chromeos_config()
+        arm_device = self._select_best_chromeos_image(devices)
+        return arm_device['version']
```
mediaminister: If there was no  Chrome OS image found, we get `NoneType' object has no attribute '__getitem__`

lib/inputstreamhelper.py
```diff
@@ -519,10 +559,10 @@ def _install_widevine_x86(self):
     def _install_widevine_arm(self):
         """Installs Widevine CDM on ARM-based architectures."""
         root_cmds = ['mount', 'umount', 'losetup', 'modprobe']
-        cos_config = self._chromeos_config()
-        device = [x for x in cos_config if config.CHROMEOS_ARM_HWID in x['hwidmatch']][0]
-        required_diskspace = int(device['filesize']) + int(device['zipfilesize'])
-        if xbmcgui.Dialog().yesno(LANGUAGE(30001),
+        devices = self._chromeos_config()
+        arm_device = self._select_best_chromeos_image(devices)
+        required_diskspace = int(arm_device['filesize']) + int(arm_device['zipfilesize'])
```
mediaminister: It's possible that `arm_device` is `None`, so we better exit this function before doing stuff with a non existing object.

lib/inputstreamhelper.py
```diff
@@ -630,7 +670,7 @@ def _update_widevine(self):
             current_version = wv_config['version']
         else:
             component = 'Chrome OS'
-            current_version = [x for x in wv_config if config.CHROMEOS_ARM_HWID in x['hwidmatch']][0]['version']
+            current_version = self._select_best_chromeos_image(wv_config)['version']
```
mediaminister: Not sure this makes sense if `wv_config` is a single object.

lib/inputstreamhelper.py
dagwieers: Yes, still not infallible yet. Working on that ;-)

dagwieers: Fixed.

lib/inputstreamhelper.py
dagwieers: Correct, at the moment there is no support for the case where there is no match. And we probably should, however. Before this change we only selected on a single HWID, without error-handling. And now we have the possibility of 20 supported HWIDs.

So I think this PR improves it to the point that not finding a matching HWID is improbable. But I agree we should take into account other causes (like the file being corrupt, empty, etc.)

lib/inputstreamhelper.py
dagwieers: Agreed. But that was an issue before as well.

dagwieers:
> I feel like part of the code is pretty self explanatory which makes some of the comments redundant. 

That's always a difficult one. I agree, but then again, I wrote it a few hours ago so it's still fresh :-)

> Too bad there's no way to scrape ARM devices from recovery.conf. It would have been nice not having to maintain this list ourselves.

I agree. The only possibility now is to look inside the images one by one :-( I actually looked at the device name and searched on Google what CPU it was using :-/

Now, the aim is to test this as part of the integration tests, and I would like to report changes based on this. So I would like to report every time a different HWID is select, or the image size changed, or the version changed. In **bold** and flashy colors, so that we are very aware when this is impacting our users as well.

And then we also need to think about revising this list every X months to stay on top of things. However as older devices disappear from the supported list, the size will increase :-(

Also, my first implementation was using only the **zipfilesize**. I figured that a small zip file size would result in a smaller image as well. But that turned out a wrong expectation, if you only take the **zipfilesize** into account, you end up with `JAQ` and a disk space requirement of 2.7GB instead of 2.0GB (and almost double the time to uncompress IIRC).

lib/inputstreamhelper.py
dagwieers: This shouldn't happen, because we do not get to this point if we couldn't get a version from recovery.conf (using the same mechanism).

lib/inputstreamhelper.py
dagwieers: wv_config (in the ARM context) is a JSON listing.

dagwieers:
This is the current decision tree:
```
[script.module.inputstreamhelper-0.3.5]: Find best ARM image to use from recovery.conf
[script.module.inputstreamhelper-0.3.5]: SKATE (2128801384) is smaller than BOB (3373181067)
[script.module.inputstreamhelper-0.3.5]: ELM (12239.67.0) is newer than SKATE (12105.100.0)
[script.module.inputstreamhelper-0.3.5]: HANA (3373917985) is smaller than ELM (3374763553)
[script.module.inputstreamhelper-0.3.5]: BIG (2139898279) is smaller than HANA (3373917985)
```

dagwieers:
So while testing I often trigger a bug in **inpustream.adaptive** that crashes Kodi. This seems to happen when you trigger an installation of Widevine (at least) twice in the same session when clicking on a DRM stream.

So something in **inputstream.adaptive** related to loading/using Widevine is causing issues on subsequent updates (even though the library is updated with itself). I think same code path was not designed to be run more than once in a single runtime/session.

@peak3d

PR #88, opened 2019-08-21
Implement preliminary integration tests

### LINKED ISSUE #63: Add unit and integration tests

dagwieers:
This PR includes:
- Added stub kodi modules
- Added sample config files
- Integrate with Travis

This fixes #63

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@
 .tox/
 __pycache__/
 *.py[cod]
+test/cdm/
+test/userdata/addon_settings.json
+test/userdata/cdm/
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,4 +19,4 @@ script:
 - pylint lib/ test/
 #- kodi-addon-checker . --branch=krypton
 #- kodi-addon-checker . --branch=leia
-#- python test/test_ishelper.py
+- python -m unittest discover
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ all: test zip
 
 package: zip
 
-test: sanity unit run
+test: sanity unit
 
 sanity: tox pylint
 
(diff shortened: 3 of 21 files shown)


dagwieers:
This is the first implementation. It goes through a single `check_inputstream()` call.
Unfortunately, this is only x86 at this time, and only a single code path is being tested.
I fixed a few genuine issues with our code during writing these tests and stub code!

dagwieers:
Also, I had to include kodi-six, because it is not available on PIP.

dagwieers:
@emilsvennesson Can you disable one of the Travis CI runs? I am not sure where this is coming from though.

vlmaksime:
@dagwieers, try to replace kodi-six with Kodistubs from my branch [Kodistubs/develop](https://github.com/vlmaksime/Kodistubs/tree/develop)
It's can emulate Kodi modules actions and easy to install with command
`  - pip install https://github.com/romanvm/script.module.simpleplugin/archive/develop.zip`
Here is an example of using Kodistubs in my addon test[plugin.video.tvzavr.ru/develop](https://github.com/vlmaksime/plugin.video.tvzavr.ru/tree/develop)

dagwieers:
@vlmaksime Kodistubs and kodi-six are two different things. So replacing kodi-six is not an option.
Also Kodistubs doesn't do a lot of things we want to do, so I don't see a good use of Kodistubs ATM.

vlmaksime:
> Also Kodistubs doesn't do a lot of things we want to do

In my branch there is a fully functional Addon class and parts of other classes that were used in my add-ons.
If you do not mind, I can add parts of code from your kodi modules into my branch. 
Also, You can replase kodi-six modules with Kodistubs modules using Mock library.

dagwieers:
@vlmaksime kodi-six is used for real, not just for testing, so there is no point in Mock.
For the stubs, we don't need a giant replacement, just the necessary stuff to run these tests.
And different modules probably have different needs wrt. stubs. So there is no point IMO to share.
It all depends to what you are really testing, and that's very context-sensitive.

dagwieers:
I am going to go forward with this. This will be an ongoing effort in adding more integration and unit-tests, especially when we add new functionality.

At some point in the future I hope we can get rid of kodi-six in this repo as well, either because it is in PIP or because Python 2.7 becomes legacy.

Issue #77, opened 2019-08-19
Stuck on old widevine after canceling first run hotfix


cache-sk:
When user hit cancel on forced widevine update on first run ("i don't have time for this now"), he stays on old widevine.
Workaround for this is manual edit "version" in settings.xml to "0" (for example).

dagwieers:
From an implementation point of view this makes sense, there is only one first-run detection. A second run would not be a first run ;-) If we wouldn't do it like this, it would have to check every time which defeats the purpose.

We would like to have a settings menu with the option to upgrade, downgrade, remove or select a CDM (planned for v0.5.0), but that means that add-ons have to add a reference to this settings menu because a script/module plugin is not listed in your add-on list, so users would not be able to find the inputstreamhelper settings :-(

We still need to have a discussion related to this. I don't see a good option in general because of the way this mechanism works.

dagwieers:
In the past the `WIDEVINE_UPDATE_INTERVAL_DAYS` was set to 1 day, which looks perfect, but actually means that every user will have a Widevine update the same day it is released (within 24h).

And if a broken/incompatible Widevine would be released (which happened before) all users are immediately impacted. The previous value of `WIDEVINE_UPDATE_INTERVAL_DAYS` was 30 days (so roughly 3% of our users would get an update, every day), which is potentially problematic as the advance warning for CDM revocation was less than 45 days in July. If that falls below our interval, our users would be affected too. That is why we decided to turn it down to 14 days, about 7% of our users will be impacted per day.

dagwieers:
Closed by #91

Issue #54, opened 2019-04-06
How to use inputstreamhelper from PVR Addon


flubshi:
I have already included inputstreamhelper in a python video addon. This works pretty good. Especially the widevine download improves the user experience a lot.

I am also working on a PVR plugin. This needs widevine as well and it would be great to use inputstreamhelper for the setup. As the PVR plugins are written in C++, I am a bit helpless how to call check_inputstream() or how to use a python library from there...

My current idea is to extend PR #52 created by @horstle such that a PVR plugin could execute this, e.g. by RunScript or the JSON API.

Or is there already a solution to use inputstreamhelper from C++/PVR Plugins?

flubshi:
A short update to explain my current workaround: I added the inputstreamhelper as dependency. 
A small python script is shipped with the c++ pvr addon. The python script calls check_inputstream(). The python helper script is linked to plugin preferences. A big disadvantage is, that a user has to run the script manually...
I would prefer to implement a cleaner solution.

dagwieers:
I would like to add RunPlugin API calls for common needs by other add-ons.
https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/59#issuecomment-521318459

dagwieers:
We still need to add check_inputstream() support to the API.

dagwieers:
Now #98 implements this.

PR #52, opened 2019-03-23
Add options to run manually and force a reinstall via the settings page.

horstle:
Hi,

I did not test this yet, but since it was requested multiple times (#47 and #32) I thought I'd create this draft PR so others could have a look at it/ test it.

Issue #47, opened 2019-03-09
Manually start libwidevine update


joendres:
I'm not sure if this is a support question or an enhancement suggestion: I'd like to start a libwidevine update manually.

What happened:
I got a message box about a libwidevine update and started it. It didn't work because ther was not enough free disk space. I added disk space, but now I don't know how to trigger the update again. Just starting a video (as the first time I got the message) doesn't work.

horstle:
Currently the easiest way to force a reinstall is to remove the directory "userdata/addon_data/script.module.inputstreamhelper". In Libreelec that's found under /storage/.kodi.
I had been working on this a few months ago, I'll work on it a bit further and probably create a pull request in the next few days.

emilsvennesson:
Yeah, this is definitely something that should be made possible.

I have been quite busy the last couple of months and haven't had the time or effort to work or review any pull requests, but I'm going to continue working on this module next month. Sorry for going MIA!

AreaScout:
you can set ``WIDEVINE_UPDATE_INTERVAL_DAYS = 0`` in config.py, then it will check on each movie you play

joendres:
@horstle Thanks for the hint. I was able to force an update by removing the files from that directory.

joendres:
@AreaScout Thanks for the hint. But since fully automatic installation doesn't work for me, it doesn't solve my problem.

dagwieers:
@joendres We are revising how we would like the plugin to behave, and what control we can put in the hands of the users #85, insights from users are very much appreciated.

dagwieers:
Closed by #91

PR #91, opened 2019-08-22
Implement settings menu and API

### LINKED ISSUE #32: Option to disable inputstreamhelper in settings + manual download cdm over GUI settings

dagwieers:
This PR includes:
- A Troubleshooting category with the option to disable the plugin
- Add an item to install Widevine CDM
- Add an item to remove Widevine CDM
- Add an API (addon.py) using script.module.routing

This fixes #32, #47, #52, #54 and #77

### DIFF
--- a/addon.xml
+++ b/addon.xml
@@ -2,10 +2,18 @@
 <addon id="script.module.inputstreamhelper" version="0.3.5" name="InputStream Helper" provider-name="emilsvennesson">
     <requires>
         <import addon="xbmc.python" version="2.25.0"/>
+        <import addon="script.module.inputstreamhelper" version="0.3.5" optional="true"/>
         <import addon="script.module.kodi-six" version="0.0.4"/>
         <import addon="script.module.pysocks" version="1.6.8" optional="true"/>
+        <import addon="script.module.routing" version="0.2.0"/>
     </requires>
-    <extension library="lib" point="xbmc.python.module"/>
+    <extension point="xbmc.python.module" library="lib"/>
+    <!-- This is needed for running from RunPlugin() -->
+    <extension point="xbmc.python.pluginsource" library="addon.py">
+        <provides>executable</provides>
+    </extension>
+    <!-- This is needed for running from Information pane -->
+    <extension point="xbmc.python.script" library="addon.py"/>
     <extension point="xbmc.addon.metadata">
         <description lang="en_GB">A simple Kodi module that makes life easier for add-on developers relying on InputStream based add-ons and DRM playback.</description>
         <description lang="el_GR">Μία απλή μονάδα για το Kodi η οποία διευκολύνει την ζωή των προγραμματιστών οι οποίοι εξαρτώνται από τα πρόσθετσ InputStream και αναπαραγωγή τύπου DRM.</description>
--- a/resources/settings.xml
+++ b/resources/settings.xml
@@ -1,5 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <settings>
-   <setting label=30053 id="disabled" type="bool" value="" />
-   <setting id="last_update" value="" visible="false" />
+    <setting id="last_update" value="" visible="false" />
+    <setting id="version" value="" visible="false" />
+    <category label="30900"> <!-- Troubleshooting -->
+        <setting label="30903" help="30904" type="bool" id="disabled" value="false"/>
+        <setting label="30904" type="text" enable="false"/> <!-- disabled_warning -->
+        <setting label="30905" help="30906" type="action" id="install_widevine" action="RunPlugin(plugin://script.module.inputstreamhelper/widevine/install)" enable="eq(-2,false)" visible="!system.platform.android"/>
+        <setting label="30907" help="30908" type="action" id="remove_widevine" action="RunPlugin(plugin://script.module.inputstreamhelper/widevine/remove)" enable="eq(-3,false)" visible="!system.platform.android"/>
+    </category>
 </settings>
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ git_branch = $(shell git rev-parse --abbrev-ref HEAD)
 git_hash = $(shell git rev-parse --short HEAD)
 
 zip_name = $(name)-$(version)-$(git_branch)-$(git_hash).zip
-include_files = addon.xml changelog.txt LICENSE.txt README.md lib/ resources/
+include_files = addon.py addon.xml changelog.txt LICENSE.txt README.md lib/ resources/
 include_paths = $(patsubst %,$(name)/%,$(include_files))
 exclude_files = \*.new \*.orig \*.pyc \*.pyo
 zip_dir = $(name)/
--- a/addon.py
+++ b/addon.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+''' This is the actual InputStream Helper plugin entry point '''
+
+from __future__ import absolute_import, division, unicode_literals
+import sys
+from routing import Plugin
+from inputstreamhelper import install_widevine, open_settings, remove_widevine
+
+plugin = Plugin()
+
+
+@plugin.route('addon.py')  # This is the entry point from the addon menu
+@plugin.route('/')
+@plugin.route('/settings')
+def settings():
+    ''' Entry point to open the plugin settings '''
+    open_settings()
+
+
+@plugin.route('/widevine/install')
+@plugin.route('/widevine/install/latest')
+def widevine_install():
+    ''' The API interface to install Widevine CDM '''
+    install_widevine()
+
+
+@plugin.route('/widevine/remove')
+def widevine_remove():
+    ''' The API interface to remove Widevine CDMs '''
+    remove_widevine()
+
+
+if __name__ == '__main__':
+    plugin.run(sys.argv)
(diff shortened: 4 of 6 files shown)


dagwieers:
cc @horstle

dagwieers:
cc @horstle @mediaminister @basrieter @emilsvennesson 

This is almost done. It adds and API to remove and install Widevine CDM.

In the future this can also be used to add an API for selecting a specific version or rolling back to a previous version of Widevine CDM. Please try it and provide feedback.

The nice thing about this is that we now have a proper addon-icon for users to access the settings directly. However I do expect that users will probably enter the settings through add-on settings references.

dagwieers:
So I am not sure about the "disable" option, I still feel it should be "enabled" and "true" by default.
Now if you "enable" "disabled", the option to install/remove become disabled. Weird IMO.

dagwieers:
@mediaminister If we have this added to inputstreamhelper, we can effectively remove our own handler from VRT NU. And everything falls into place ;-)

horstle:
I think it would be better if it stayed "invisible". Most of the time users won't care about this and it would only be in their way, but if something fails, they would probably first look at the settings of the corresponding addon, so it's critical to get added there. Maybe at least in the inputstream.adaptive settings, since they are already referenced by some.

About the "disabled". If you really want to change this, I would suggest creating another setting "enabled" and hiding "disabled" from the settings page, then check for both settings and show a warning that "disabled" will be deprecated, if it is set. A few releases later you could then remove the "disabled" setting.

horstle:
I think "disable" (or "enable") shouldn't be in the "Troubleshooting" category, but rather in "Configuration" (as you suggested in https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/85#issue-482954477) or just be the very first one.

dagwieers:
> I think it would be better if it stayed "invisible". 

We cannot make it invisible. The icon will appear. We can however make it do nothing.

> I think "disable" (or "enable") shouldn't be in the "Troubleshooting" category, but rather in "Configuration"

I disagree now. Everything in the settings are for troubleshooting.. The plugin should work out of the box and we should not expect people to have to use this.

That is why everything belongs together under Troubleshooting. And the upside is that we can disable everything if the plugin is disabled.

addon.xml
```diff
@@ -2,10 +2,18 @@
 <addon id="script.module.inputstreamhelper" version="0.3.5" name="InputStream Helper" provider-name="emilsvennesson">
     <requires>
         <import addon="xbmc.python" version="2.25.0"/>
+        <import addon="script.module.inputstreamhelper" version="0.3.5" optional="true"/>
```
emilsvennesson: Why is this needed? :frowning:

horstle:
> We cannot make it invisible. The icon will appear. We can however make it do nothing.
Cluttering the Addon section with an icon that does nothing would make no sense at all. If it has to be there, it might as well have a function.

horstle:
> I disagree now. Everything in the settings are for troubleshooting.. The plugin should work out of the box and we should not expect people to have to use this.
> 
> That is why everything belongs together under Troubleshooting. And the upside is that we can disable everything if the plugin is disabled.

Once this is merged I'll rebase and fix #21 again. Should the setting for a custom path then be in the Troubleshooting section, too?

resources/settings.xml
```diff
@@ -1,5 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <settings>
-   <setting label=30053 id="disabled" type="bool" value="" />
-   <setting id="last_update" value="" visible="false" />
+    <setting id="last_update" value="" visible="false" />
+    <setting id="version" value="" visible="false" />
+    <category label="30900"> <!-- Troubleshooting -->
+        <setting label="30903" help="30904" type="bool" id="disabled" value="false"/>
+        <setting label="30904" type="text" enable="false"/> <!-- disabled_warning -->
```
mediaminister: The #30904 string is too long to show on one line in the settings dialog, it breaks at `playback ma...`

addon.xml
dagwieers: Otherwise our "plugin" or "script" would not be able to import inputstreamhelper.
It looks silly, but it works.

resources/settings.xml
dagwieers: It worked for my setup at home and my celphone, but I was worried this, yes.
The real solution is to use the new v18 settings syntax which can show help statements.

dagwieers:
> Perhaps we should split the API code into its own file? inputstreamhelper.py is starting to become quite complex, and it might be time to start separating some of the code.

I think we should, but can we postpone this until after v0.4.0 is released?
Can you create a ticket to discuss this?

dagwieers:
Let's merge this now, and create PRs for the individual issues that were raised.
(I am off this weekend, so if this is merged, others can start building on this)

mediaminister:
@emilsvennesson @dagwieers Removing Widevine CDM doesn't work on Windows. I have a fix for this, but I would like to clean up all Widevine CDM file operations.
But before I continue I would like to now why we use symlinks from the inpustreamhelper addon data dir and the Kodi CDM dir? Is this necessary? Is there a specific reason for this? Can I safely remove this to drastically simplify Widevine CDM file operations.

horstle:
afaik the libwidevinecdm.so is stored in the inputstreamhelper data dir, but inputstream.adaptive (obviously) looks for it in the CDM dir. So a symlink seems the obvious solution. On Windows instead of a symlink the lib is actually copied instead of linked. That's probably why the removal doesn't work.

mediaminister:
The removal didn't work because Windows uses `widevinecdm` and not `libwidevinecdm` as filename: https://github.com/peak3d/inputstream.adaptive/blob/980a42053c3fa0fa305d7349d4c74af50479f249/wvdecrypter/CMakeLists.txt#L41-L50

As a test I disabled inputstreamhelper and manually removed  inputstreamhelper cdm data dir.
DRM content just plays fine, as I expected because InputStream Adaptive always looks in the Kodi CDM dir:

https://github.com/peak3d/inputstream.adaptive/blob/80e4dbf147529f9edbd06a5b8b958c31666308da/CMakeLists.txt#L86-L90

My conclusion is that symlinking or copying `from` Kodi CDM dir to InputstreamHelper data dir is useless. It doesn't make sense to me.

horstle:
> The removal didn't work because Windows uses `widevinecdm` and not `libwidevinecdm` as a filename:

Ok, didn't look at the code. Then that.

> As a test I disabled inputstreamhelper and manually removed inputstreamhelper cdm data dir.
> DRM content just plays fine, as I expected because InputStream Adaptive always looks in the Kodi CDM dir:

That's what I thought was the problem, but on linux that would have broken inputstream.adaptive, since the symlink in CDM dir would still exist, but not its target.

> My conclusion is that symlinking or copying **from** Kodi CDM dir **to** InputstreamHelper data dir is useless. It doesn't make sense to me.

I agree, copying from CDM to data would make no sense, but I couldn't find any code which would do that.
About the symlinks, a symlink in the CDM dir targeting the addon data dir (that's how I interpret "**from** Kodi CDM dir **to** InputstreamHelper data dir" in that case)  does make sense. However to create such a symlink with `os.symlink()` you have to specify the addon data dir first and then the CDM dir (a bit counter intuitive, but makes it consistent with `copy()`). Maybe that's where this issue comes from.

mediaminister:
I wonder why we don't use the Kodi CDM folder to install the Widevine CDM. Symlinking from one directory to another directory makes the workflow more complicated. There must be reason for it, otherwise we should just use the Kodi CDM folder to install and remove Widevine CDM.

emilsvennesson:
The reasoning behind this can be found in #1:

> @emilsvennesson Can't find the license where I read the part about extracting/copying it somewhere is forbidden. I did lots of research regarding available sources (legal ones) for ARM versions of widevine & I came to the conclusion that downloading the chromeos image & extracting it from there might be the best shot for not making Google angry.
I´d also asked friends at Mozilla & Google if there´s another way to get it for the development purposes only & they didn't know of any.

> My suggestion would be to use the ChromeOS image, symlink the libwidevine & delete everything      but the original path to the extracted version. Guess that is as close we can get to have a solution that might not violate the license.

_Originally posted by @asciidisco in https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/1#issuecomment-329738013_

I don't think this is necessary though. I have yet to find anywhere in the Widevine license that states that it's forbidden to copy the binary (and we're already doing it on Windows), so this part of the code can definitely be simplified. It has just been there from the start and I've neglected to change it in favor of other things.

mediaminister:
I completely agree with  @asciidisco 
A Kodi user already **copies** the software from a Google server to their hard disk when accepting the EULA. I think its better that this copy ends up in the right directory with as little as possible operations. In my opinion there is no problem as long as the software stays on the users (private) hard disk.
For instance, we should never upload copyrighted software to cloud services, even a (copyrighted) EULA or configuration file should never end up on Github.

dagwieers:
So the required fix in **script.module.routing** (tamland/kodi-plugin-routing#16) has been accepted. We are now waiting for a new release.

PR #95, opened 2019-08-24
directly remove and install Widevine in Kodi CDM dir

mediaminister:
This pull request includes:
- fix for removing Widevine on Windows
- move `remove_widevine` to Helper class
- always use Kodi CDM dir to install Widevine CDM to, phase out use of inputstreamhelper addon data CDM dir
- determine widevine path based on static filename (just like InputStream Adaptive does)
- wait until InputStream Adaptive installation is complete (xmbc.executebuiltin wait parameter)
- use notification dialogs when installing or removing Widevine CDM

### DIFF
--- a/lib/config.py
+++ b/lib/config.py
@@ -15,11 +15,12 @@ DRM_SCHEMES = {
     'com.widevine.alpha': 'widevine'
 }
 
-CDM_EXTENSIONS = (
-    '.so',
-    '.dll',
-    '.dylib'
-)
+WIDEVINE_CDM_FILENAME = {
+    'Android': None,
+    'Linux': 'libwidevinecdm.so',
+    'Windows': 'widevinecdm.dll',
+    'Darwin': 'libwidevinecdm.dylib'
+}
 
 ARCH_MAP = {
     'aarch64': 'arm64',
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -76,31 +76,6 @@ def system_os():
     return system_os.name
 
 
-def open_settings():
-    ''' API entry point for opening inpustreamhelper settings '''
-    ADDON.openSettings()
-
-
-def install_widevine():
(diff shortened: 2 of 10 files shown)


dagwieers:
This is how it should have been. Thanks!

lib/inputstreamhelper.py
```diff
@@ -88,17 +88,7 @@ def install_widevine():
...
-    for extension in config.CDM_EXTENSIONS:
-        widevinecdm = os.path.join(cdm_path, 'libwidevinecdm' + extension)
-        if os.path.exists(widevinecdm):
-            log('Remove Widevine CDM at {0}'.format(widevinecdm))
-            os.remove(widevinecdm)
-            xbmcgui.Dialog().ok('{0} v{1}'.format(ADDON_ID, ADDON_VERSION), LANGUAGE(30051))
-            break
-    else:
-        xbmcgui.Dialog().ok('{0} v{1}'.format(ADDON_ID, ADDON_VERSION), LANGUAGE(30052))
+    Helper('mpd', drm='widevine').remove_widevine()
```
dagwieers: Maybe you call this directly from addon.py, like I am doing in #98

lib/config.py
```diff
@@ -15,11 +15,12 @@
     'com.widevine.alpha': 'widevine'
 }
 
-CDM_EXTENSIONS = (
-    '.so',
-    '.dll',
-    '.dylib'
-)
+WIDEVINE_CDM_FILENAME = {
+    'Android': 'libwidevinecdm.so',
```
dagwieers: Shouldn't Android be `None`? As we don't manage it there.
The API should just skip any possible removal.

mediaminister: Indeed.

lib/inputstreamhelper.py
mediaminister: Okay, I rebased and fixed this.

PR #96, opened 2019-08-24
use local url variable

mediaminister:


dagwieers:
Yes, much better. I didn't want to change too much when moving from requests to urllib.

PR #97, opened 2019-08-24
Dutch translation fixes

dagwieers:
These are older fixes that I saved for when we were going to rework the
dialogs. But maybe it's better to do it now.

### DIFF
--- a/resources/language/resource.language.nl_nl/strings.po
+++ b/resources/language/resource.language.nl_nl/strings.po
@@ -19,7 +19,7 @@ msgstr "Informatie"
 
 msgctxt "#30002"
 msgid "This add-on relies on the proprietary decryption module [B]Widevine CDM[/B] for playback."
-msgstr "Deze add-on heeft een decryptie-module van een derde partij nodig [B]Widevine CDM[/B] om af te spelen."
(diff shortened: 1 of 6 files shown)


resources/language/resource.language.nl_nl/strings.po
```diff
@@ -19,22 +19,22 @@ msgstr "Informatie"
 
 msgctxt "#30002"
 msgid "This add-on relies on the proprietary decryption module [B]Widevine CDM[/B] for playback."
-msgstr "Deze add-on heeft een decryptie-module van een derde partij nodig [B]Widevine CDM[/B] om af te spelen."
+msgstr "Deze add-on heeft de decryptie-module [B]Widevine CDM[/B] van een derde partij nodig om af te spelen."
```
mediaminister: "propriëtaire software" is also a Dutch term, so "propriëtaire decryptiemodule" is also possible and shorter.

dagwieers: Yeah, but I don't know if people would understand it if we did use it. I am also not sure if this is used in the Netherlands. Personally, I wouldn't add that information in the dialog at all, but I don't know if we should.

mediaminister: Okay, we'll see later on which messages are needed and which not.

Issue #92, opened 2019-08-22
UnicodeError when installing Widevine CDM on ARM


dagwieers:
While testing the new API, I got this:
```log
2019-08-23 01:21:38.917 T:1207174000   ERROR: /storage/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper.py:759: UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal
                                              if filename == 'libwidevinecdm.so':
```

indritix:
I can confirm the UnicodeError on my rpi3 (LE)
`2019-08-24 09:01:01.316 T:1575908224   ERROR: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                             - NOTE: IGNORING THIS CAN LEAD TO MEMORY LEAKS!
                                            Error Type: <type 'exceptions.UnicodeDecodeError'>
                                            Error Contents: 'ascii' codec can't decode byte 0xc5 in position 29: ordinal not in range(128)
                                            Traceback (most recent call last):
                                              File "/storage/.kodi/addons/script.module.inputstreamhelper/addon.py", line 34, in <module>
                                                plugin.run(sys.argv)
                                              File "/storage/.kodi/addons/script.module.routing/lib/routing.py", line 115, in run
                                                self._dispatch(path)
                                              File "/storage/.kodi/addons/script.module.routing/lib/routing.py", line 126, in _dispatch
                                                view_func(**kwargs)
                                              File "/storage/.kodi/addons/script.module.inputstreamhelper/addon.py", line 24, in widevine_install
                                                install_widevine()
                                              File "/storage/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper.py", line 86, in install_widevi
ne
                                                Helper('mpd', drm='widevine').install_widevine()
                                              File "/storage/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper.py", line 694, in install_widev
ine
                                                return self._install_widevine_arm()
                                              File "/storage/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper.py", line 665, in _install_wide
vine_arm
                                                self._extract_widevine_from_img()
                                              File "/storage/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper.py", line 766, in _extract_wide
vine_from_img
[...]

PR #99, opened 2019-08-26
force bytestring for os.walk

mediaminister:
This forces bytestrings for python 2 os.walk, related to https://bugs.python.org/issue22862

### DIFF
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -748,9 +748,9 @@ class Helper:
 
     def _extract_widevine_from_img(self):
         """Extracts the Widevine CDM binary from the mounted Chrome OS image."""
-        for root, dirs, files in os.walk(self._mnt_path()):  # pylint: disable=unused-variable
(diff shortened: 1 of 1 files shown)


dagwieers:
This fixes #92 ?

mediaminister:
Yes, as long as `libwidevinecdm.so` doesn't contain special chars.
For a next version we should think about removing `kodi_six` and handling unicode in a more transparent way. Because now `unicode_literals` and `kodi_six` are doing stuff and in the end we don't have a clear overview anymore of what is happening with text processing.

Maybe using our own `from_unicode` and `to_unicode` functions is a better idea. https://github.com/pietje666/plugin.video.vrt.nu/blob/8cd3304f859e2febaf9f89be91a0da3cc3ed329d/resources/lib/statichelper.py#L71-L81

dagwieers:
> For a next version we should think about removing kodi_six and handling unicode in a more transparent way.

Can you open a ticket for this? (So that we have some more diversity in who creates tickets ;-))

PR #101, opened 2019-08-27
Add a widevine_reinstall API call

dagwieers:
This adds a widevine_reinstall API call, which basically calls
remove_widevine() and install_widevine().

This was an idea coming from @basrieter

### DIFF
--- a/addon.py
+++ b/addon.py
@@ -36,5 +36,14 @@ def widevine_remove():
     Helper('mpd', drm='widevine').remove_widevine()
 
 
+@plugin.route('/widevine/reinstall')
+@plugin.route('/widevine/reinstall/latest')
+def widevine_reinstall():
+    ''' The API interface to reinstall Widevine CDM '''
+    is_helper = Helper('mpd', drm='widevine')
+    is_helper.remove_widevine()
+    is_helper.install_widevine()
+
+
 if __name__ == '__main__':
     plugin.run(sys.argv)
--- a/test/test_routing.py
+++ b/test/test_routing.py
@@ -23,12 +23,15 @@ class TestRouter(unittest.TestCase):
 
     def test_widevine_install(self):
         plugin.run(['plugin://script.module.inputstreamhelper/widevine/install', '0', ''])
-        plugin.run(['plugin://script.module.inputstreamhelper/widevine/install/latest', '0', ''])
         self.assertEqual(plugin.url_for(addon.widevine_install), 'plugin://script.module.inputstreamhelper/widevine/install/latest')
 
     def test_widevine_remove(self):
-        plugin.run(['plugin://script.module.inputstreamhelper', '0', ''])
+        plugin.run(['plugin://script.module.inputstreamhelper/widevine/remove', '0', ''])
         self.assertEqual(plugin.url_for(addon.widevine_remove), 'plugin://script.module.inputstreamhelper/widevine/remove')
+
+    def test_widevine_reinstall(self):
+        plugin.run(['plugin://script.module.inputstreamhelper/widevine/reinstall', '0', ''])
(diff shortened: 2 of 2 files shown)


mediaminister:
Looks good to me, but what is the purpose of the separate install and remove functions? Is this useful? The remove method just clears the cdm file and the two different install methods are also checking the current installed version and removing things.

dagwieers:
Good question, it was useful before we had the reinstall option. But the reinstall option 8s not in the settings currently.

dagwieers:
I wonder, should we only have the "Install Widevine CDM" option in the settings that actually does a removal and installation (i.e. reinstallation)? So dumb things down a bit from the user-perspective.

I am undecided.

mediaminister:
I think we need the following changes:

1) `remove_widevine` should delete all widevine files
2) `install_widevine` should call remove_widevine before before a new install

After these changes, we need only need an install setting, we can keep an api call for removing widevine for add-ons or users who want to disable DRM completely.

dagwieers:
So "Install Widevine CDM" in the settings (and API) will always remove the existing Widevine CDM.
Potentially break a working situation? Or should we specifically call it "'(Re)install Widevine CDM".

Also, I envisioned add-ons linking to the inputstreamhelper settings (for troubleshooting, like we do in VRT NU), but if it always reinstalls I am not sure if that is a good idea.

So what relation do we want with addons using InputStream Helper?
1. Should they all individually be adding actions to install/reinstall Widevine CDM from their addons?
2. Should they add a link to the InputStream Helper addon settings?
3. Should they not have any reference to InputStream Helper (aside from calling it when playing content)?

I assumed (2) but I am not sure if this is what others actually expect. How would users end up where they need to end up in the face of Widevine issues, and what tools do they need.

*Obviously we remember how it used to work, and the issues we used to have, but this is probably more about what issues do we expect in the future, and how will our future users be undergoing these issue.*

mediaminister:
2 is the most logical option.

> So "Install Widevine CDM" in the settings (and API) will always remove the existing Widevine CDM.

I do not see another option, installing something new always means possibly breaking the current situation, right?

dagwieers:
What I meant was that "Install Widevine CDM" currently would not install it if it was already installed. The fallacy here obviously is that this check doesn't know whether it actually worked, and immediately reports that is was successfully installed (where it did not do anything really). This we *should* fix, I'll open a new issue for this specifically.

But having a distinction between "Install" and "Reinstall" is one that add-on developers may expect to have. As they don't know whether it currently is already installed, and/or it actually works. So they may want to have a "Install Widevine CDM" option in their settings that does not reinstall needlessly.

That is mostly why I figured having all (API) options is a better deal, and only having Remove and Install in the settings offers every possible permutation from the settings already.

horstle:
Wouldn't `install_widevine()` (i.e. the `widevine_install()` API call) actually overwrite any existing files? I'm not at home until friday, so I can't test this atm, but from what I can see looking at the code on my phone (might have overlooked something), there are no checks for existing files.

basrieter:
I would say only have the _install_ and _remove_ api. Then if you _install_ and it is already there, show a warning. 

Also an idea is to keep an old version as backup? so they can roll back?

dagwieers:
@basrieter Backup and rollback (#36) is something we plan for v0.5.0. We want to release as quickly as possible now. But those will be part of that same (Troubleshooting) settings menu.

dagwieers:
@horstle You are correct, the widevine_install API replaces any existing Widevine already. It was our own addition in VRT NU that is using `check_install()` that was in fact testing and reporting success even though it was installed. So the re-install addition has no value.

I am sorry for bringing this confusion to you all.

PR #119, opened 2019-08-29
use "lsep" type for label instead "text"

vlmaksime:


dagwieers:
Why `lsep` instead of `text`?

vlmaksime:
"lsep" allows you to display long lines, when the "text" trims the end. Here is example: [Imgur](https://imgur.com/9IJ5LOK)

dagwieers:
I understand, but `lsep` is used as line-separator. So visually it marks a new category with a category title. And that is not what we want here. In the future we can use actual help-messages for every menu item, but this has only been adding in Kodi Leia (v18) and is still a bit experimental.

So I don't think `lsep` is a good fit for this.

vlmaksime:
We can use "visible" attribute for hiding this text, while add-on enabled. But this feature broken on Leia [Kodi Forum](https://forum.kodi.tv/showthread.php?tid=346586)

dagwieers:
I agree there is a problem, and we coped with it by making the text as short as possible.
But I don't like this solution. Maybe this ought to be fixed in Kodi/skin instead.

vlmaksime:
Ok

dagwieers:
@vlmaksime Can you open an issue for this (incl.. a screenshot)? We should open an issue upstream and track this issue here, even if it is in Kodi or in the skin.

vlmaksime:
Or I can make shorter translation for this string.

dagwieers:
@vlmaksime That is what we did for English and Dutch (see https://github.com/emilsvennesson/script.module.inputstreamhelper/pull/91#pullrequestreview-278910899), but I think we should push for a fix for this. There is no good reason to have a text-type that is being cut off..

PR #120, opened 2019-08-29
Update German translations

flubshi:
Translate new strings to german.


dagwieers:
Thanks !

horstle:
I would suggest some minor changes. Will create a PR soon.

PR #123, opened 2019-08-30
Fix various issues found during testing

dagwieers:
This PR includes:
- Fix use of shell for modprobe (fails on RHEL7)
- Use correct order for commands wrt. options and arguments
- Use recovery.conf.json on ARM and widevine_config.json on non-ARM
- Fix an issue on Android where widevine_cdm_filename is None
- Use decode() on HTTP GET return information
- Use decode() on _run_cmd() output
- Simplify nested if-then logic (prefer flat structure, early exit)
- Add Android/ARM integration tests


dagwieers:
There are some fixes in this one that I think ought to go in v0.4.0 as well, but this is quite late in the game. The most important one to me seems:
- Handling filename None on Android (not sure if this an issue in the real world?)

dagwieers:
I removed the potentially controversial new fdisk/parted parsing.
I am going to propose that for v0.5.0 in a subsequent PR.
Everything in this PR seems rather safe or needed for v0.4.0.

dagwieers:
There seems to be an issue with unit tests being influenced/corrupted by other unit tests on Travis.
It appears that imports are not being reloaded, but reused (probably to speed things up on Travis).

PR #128, opened 2019-08-30
Add coverage support

dagwieers:
This PR adds coverage support.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,2 @@
+[run]
+source=test/test_*.py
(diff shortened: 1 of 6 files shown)


dagwieers:
@emilsvennesson Can you see if we can add some integration in GitHub?

What I really do not want to do, is add a bot that reports on coverage for every commit.
But reporting on coverage at the end (like we do now with Travis checks) would be very nice.
(i.e. does this PR improve the coverage support? And in what way.)

PR #129, opened 2019-08-30
Fix ARM processing in unittest locally

dagwieers:
This PR includes:
- Fixes to fdisk and parted parsing (required for local unittesting on Linux)

The new processing makes parsing the image more reliable and better future-proof.

### DIFF
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -257,23 +257,64 @@ class Helper:
         return addon.getAddonInfo('version')
 
     def _chromeos_offset(self, bin_path):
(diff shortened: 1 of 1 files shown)


dagwieers:
The device handling in this PR as been simplified to the point that we only need to run 2 commands using sudo: **mount** and **umount**. This makes testing ARM a lot more convenient (no longer the need to type your sudo password 5 times).

dagwieers:
@emilsvennesson And this &uarr; is what I don't like ;-)

dagwieers:
It appears that Busybox mount does not have support for `offset`.
Which means we have to follow through on using `losetup` :-(

It is probably better to invest time in a pure-python extraction routine #72

Issue #102, opened 2019-08-27
Handle unicode in a more transparent way


mediaminister:
InputStreamHelper is used by people using different languages and alphabets on a wide variety of different systems and Kodi versions. So, it's important our add-on supports all unicode characters and maintain compatibility with Python 2 and Python 3 Kodi versions for all these users.

Today we use `kodi_six` to handle unicode for the Kodi Python API and we use `unicode_literals` and use  the encoding declaration `# -*- coding: utf-8 -*-` to make sure we always use unicode throughout the add-on source code. 
But, sometimes we need to convert unicode to bytestrings and vice-versa.
For instance `os.walk`  in Python 2 which doesn't support unicode flawlessly.
Or for instance when importing or exporting text from external sources like the internet or a filesystem.

Using different external libraries handling unicode makes it very difficult to transparently and consciously handle unicode.  For a next version we should think about removing` kodi_six` and handling unicode encoding and decoding using our own Python2/3 compatible conversion functions.

We did this before in the VRT NU add-on and this helps us transparently maintaining Python2/3 and unicode support: https://github.com/pietje666/plugin.video.vrt.nu/blob/8cd3304f859e2febaf9f89be91a0da3cc3ed329d/resources/lib/statichelper.py#L71-L81

mediaminister:
I'm working on this, but to reliable test unicode support on Windows and Kodi Python 3, there are fixes needed upstream.
There is already one fix for Kodi Python 3 for Windows users with unicode characters in their user profile path: https://github.com/xbmc/xbmc/issues/16577
But now I bump into a possible bug in InputStream Adaptive...

PR #130, opened 2019-08-30
Remove kodi-six

dagwieers:
This is an early attempt to replace kodi-six with our own unicode-wrapper.

This relates to #102


dagwieers:
Everything seems to work out all right?

mediaminister:
> Everything seems to work out all right?

I don't think it will work for Björn. (`C:\Users\Björn\AppData\Roaming\Kodi\addons\script.module.inputstreamhelper\lib\inputstreamhelper.py`)

PR #133, opened 2019-08-30
Reduce extraction to mount/unmount

dagwieers:
This PR includes:
- Remove loop device handling (let the kernel handle this through mount)

This reduces the number of sudo operations from 5 to only 2.

This relates to #129

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,3 @@
 [run]
 branch = true
-include = addon.py lib/
+omit = test/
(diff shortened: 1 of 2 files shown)


dagwieers:
The device handling in this PR as been simplified to the point that we only need to run 2 commands using sudo: mount and umount. This makes testing ARM a lot more convenient (no longer the need to type your sudo password 5 times).

dagwieers:
It appears that Busybox mount does not have support for offset.
Which means we have to follow through on using losetup :-(

It is probably better to invest time in a pure-python extraction routine #72

PR #137, opened 2019-08-31
More coverage improvements

dagwieers:
More experiments with coverage.


dagwieers:
Not sure if we can improve it any further, most of the stuff that is not covered by tests are the failure scenarios usually ending up with an error box. Not sure if there's a lot of benefit to also try and induce those errors.

Issue #87, opened 2019-08-20
Implement About dialog with Inputstream/Widevine details


dagwieers:
Currently on ARM we have no clue what version of Widevine CDM version we get from the ChromeOS image. While we can do it like this:
```shell
kodi01:~ # grep -E '[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+' ~/.kodi/cdm/libwidevinecdm.so 
4.10.1440.18
4.10.1440.18
```

It would be nice if we can do this from python instead, preferably by reading the ELF structure.

horstle:
I couldn't find the version string in the output of `readelf`, but this would do the same as your command:
```python
import re

def get_lib_version(file):
    with open(file, 'rb') as f:
        x = re.search('[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+', str(f.read()))
        return(x.group(0).lstrip('0'))
```
With `file` being the path to the libwidevinecdm.so.

dagwieers:
This would indeed be the alternative implementation (but is also a liability if it would fail).
We need unit-tests for this so we can detect issues and escalate them.

dagwieers:
I verified and this works for Linux, Windows and macOS.

dagwieers:
I think this is sufficient for showing the version in an About dialog in the settings.
Then that dialog could show:
- Version of InputStream Helper
- Current Widevine version
- Latest ChromeOS version used (stored in settings)
- Detected platform/architecture
- Location of the Widevine CDM library
- A link to where to report issues on GitHub

dagwieers:
This is implemented with #139

PR #139, opened 2019-08-31
Add About category to settings page.

horstle:
Implementing #87 . Tested on Ubuntu x86_64 and Libreelec RPi3. The content of the settings is filled after the first run.

### DIFF
--- a/lib/inputstreamhelper.py
+++ b/lib/inputstreamhelper.py
@@ -256,6 +256,14 @@ class Helper:
         addon = xbmcaddon.Addon(self.inputstream_addon)
         return addon.getAddonInfo('version')
 
+    @staticmethod
+    def _get_lib_version(lib):
+        if lib:
+            with open(lib, 'rb') as f:
+                x = re.search(r'[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+', str(f.read()))
+            return x.group(0).lstrip('0')
+        return 'Not found'
+
     def _chromeos_offset(self, bin_path):
         ''' Calculate the Chrome OS start offset using fdisk or parted '''
         # Prefer parted over fdisk because it is more reliable
@@ -654,7 +662,7 @@ class Helper:
 
         return False
 
-    def _install_widevine_arm(self):
+    def _install_widevine_arm(self):  # pylint: disable=too-many-statements
         """Installs Widevine CDM on ARM-based architectures."""
         root_cmds = ['mount', 'umount', 'losetup', 'modprobe']
         devices = self._chromeos_config()
@@ -717,6 +725,7 @@ class Helper:
                     progress_dialog.update(97, line1=localize(30050))  # Finishing
                     self._cleanup()
                     if self._has_widevine():
+                        ADDON.setSetting('chromeos_version', arm_device['version'])
                         with open(self._widevine_config_path(), 'w') as config_file:
                             config_file.write(json.dumps(devices, indent=4))
                         wv_check = self._check_widevine()
@@ -1045,3 +1054,34 @@ class Helper:
             return None
 
         return dict(http=proxy_address, https=proxy_address)
+
+    def about_dialog(self):
+        """ Show an About box with useful info e.g. for bug reports"""
+        ishelper = [ADDON.getAddonInfo('version'), '(disabled)' if not ADDON.getSetting('disabled') == 'false' else '']
+        istream = [self._inputstream_version(), '(disabled)' if not self._inputstream_enabled() else '']
+        kodi_version = self._kodi_version()
+        wv_version = self._get_lib_version(self._widevine_path())
+        wv_updated = datetime.fromtimestamp(float(ADDON.getSetting('last_update'))).strftime("%Y-%m-%d %H:%M") if ADDON.getSetting('last_update') else 'Never'
+        kodi_platform = system_os()
+        arch = self._arch()
+        cdm_dir = self._ia_cdm_path()
+        issue_url = config.ISSUE_URL
+
+        kodi_info = [localize(30921, version=kodi_version),
+                     localize(30922, kodi_platform=kodi_platform, arch=arch)]
+        is_info = [localize(30931, version=ishelper[0], enabled=ishelper[1]),
+                   localize(30932, version=istream[0], enabled=istream[1])]
+        wv_info = [localize(30941, version=wv_version, updated=wv_updated),
+                   localize(30942, cdm_dir=cdm_dir)]
+        if platform == 'arm':
+            wv_info.append(localize(30943, version=ADDON.getSetting('chromeos_version')))
+
+        text = (localize(30920) + "\n - "
+                + "\n - ".join(kodi_info) + "\n\n"
+                + localize(30930) + "\n - "
+                + "\n - ".join(is_info) + "\n\n"
+                + localize(30940) + "\n - "
+                + "\n - ".join(wv_info) + "\n\n"
+                + localize(30950, issue_url=issue_url))
+
+        xbmcgui.Dialog().textviewer(localize(30909), text)
--- a/resources/settings.xml
+++ b/resources/settings.xml
@@ -1,11 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <settings>
-    <setting id="last_update" value="" visible="false" />
-    <setting id="version" value="" visible="false" />
+    <setting id="last_update" value="" visible="false"/>
+    <setting id="version" value="" visible="false"/>
+    <setting id="chromeos_version" value="" visible="false"/>
     <category label="30900"> <!-- Troubleshooting -->
         <setting label="30903" help="30904" type="bool" id="disabled" value="false"/>
         <setting label="30904" type="text" enable="false"/> <!-- disabled_warning -->
         <setting label="30905" help="30906" type="action" id="install_widevine" action="RunScript(script.module.inputstreamhelper, widevine_install)" enable="eq(-2,false)" visible="!system.platform.android"/>
         <setting label="30907" help="30908" type="action" id="remove_widevine" action="RunScript(script.module.inputstreamhelper, widevine_remove)" enable="eq(-3,false)" visible="!system.platform.android"/>
+        <setting label="30909" type="action" id="about" action="RunScript(script.module.inputstreamhelper, about)"/>
     </category>
 </settings>
--- a/test/xbmcgui.py
+++ b/test/xbmcgui.py
@@ -37,6 +37,13 @@ class Dialog:
         print('[37;100mYESNO:[35;0m [%s] [35;0m%s[0m' % (heading, line1))
         return True
 
+    @staticmethod
+    def textviewer(heading, text=None, usemono=None):
+        ''' A stub implementation for the xbmcgui Dialog class textviewer() method '''
+        heading = kodi_to_ansi(heading)
+        text = kodi_to_ansi(text)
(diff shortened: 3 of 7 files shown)


dagwieers:
Personally, I think this ought to be a pop-up box with the information.
This has a few benefits:
- Real-time information (rather than stored bits)
- Can be called from the API
- Information can be tweaked depending on platform (i.e. show ChromeOS image only on ARM)

horstle:
Should this be only one pop-up box or multiple? If multiple, how many entries for each? And should I use `dialog.ok()` or rather something different (like `select()`, `contextmenu()` or `textviewer()`)?

By the way the ChromeOS setting would have been invisible on non ARM.

dagwieers:
@horstle Can you add a screenshot so others can chime in? I also like to hear what others think about this. 

I would go for a single box with an OK button. However currently there is only 3 lines visible IIRC. So that's not good IMO.

> By the way the ChromeOS setting would have been invisible on non ARM.

Good point.

horstle:
kodi_ish_about

That's how it would look like atm. I'm already working on a dialog with 3 ok boxes:
* helper, widevine and on arm chromeos version
* platform and architecture
* cdm dir and url to issue page

dagwieers:
I don't think 3 dialog boxes makes sense.

dagwieers:
Maybe it could be useful to have a "Make screenshot" entry so people can share this information more easily. (My wish was that when you called the dialog, it would also write this information to the log in a single block)

horstle:
With `dialog.ok()` it could show only 4 lines at once, but I tried using textviewer and its not too bad, maybe with some additional formatting... 

kodi_ish_about_textviewer

dagwieers:
@horstle That is already quite good. If we would have several sections/titles we could do:

> **Kodi information**
> - Kodi version: **18.3**
> - Kodi platform: **Linux**
> - Kodi architecture: **arm**
>
> **InputStream information**
> - InputStream Helper version: **0.4.0**
> - InputStream Adaptive version: **2.3.22**
>
> **Widevine information**
> - Widevine version: **4.10.144.19**
> - ChromeOS version: **12239.67.0**
> - Widevine CDM path: **/storage/.kodi/cdm**
>
> Please report issues to:
> - https://github.com/emilsvennesson/script.module.inputstreamhelper/issues

I hope at some point we can have a custom dialog, with a proper title and 2 buttons (Screenshot, Close).

dagwieers:
If we have this, it becomes useful for add-on developers to add in their settings an entry "**Widevine information**" using `RunPlugin(plugin://script.module.inputstreamhelper/about)`

And this will help them to get better bug reports as well. Very cool!

horstle:
OK, now it looks like this:

kodi_ish_about_textviewer1
kodi_ish_about_textviewer2

dagwieers:
It is unfortunate it does not fit on one screen :-/

horstle:
I assume the number of lines per page depends on the monitor? Or would it always be 14? In that case maybe I could remove 5 lines (the 4 too many atm plus 1 for chromeos).

dagwieers:
@horstle You are right, but we could optimize on vertical screen estate too.

I was thinking it would also be nice to indicate if a plugin is disabled.

> InputStream Helper version: **0.4.0** (disabled)
> InputStream RTMP version: **2.0.3** (disabled)

dagwieers:
And include last_updated information ;-)
Sorry for being pedantic...

horstle:
Bildschirmfoto von 2019-08-31 17-46-49

last_update and disabled are in. I also removed some newlines. Now on my laptop I would have 1 line left for ChromeOS, but not for Inputstream RTMP. I could make platform and architecture one line, then it would fit, but still no idea how this looks on other displays.

dagwieers:
Looking good already.
- I would not show "(enabled)", only "(disabled)". People would assume it is enabled when it is shown, but not that it could be disabled.
- Putting Platform and architecture together seems like a good idea.
> Platform: **Linux/x86_64**

- Change the order of Widevine info:
  - First version, then path, we can save one more line doing:
> Widevine version: **4.10.1440.18** (last update on **2019-08-31 12:51**)

No need to show seconds.

horstle:
Now I actually have the whole thing on one page.

Bildschirmfoto von 2019-08-31 18-29-41

dagwieers:
Looking good.

test/xbmcgui.py
```diff
@@ -37,6 +37,13 @@ def yesno(heading, line1, line2=None, line3=None, nolabel=None, yeslabel=None, a
         print('[37;100mYESNO:[35;0m [%s] [35;0m%s[0m' % (heading, line1))
         return True
 
+    @staticmethod
+    def textviewer(heading, text=None, usemono=None):
+        ''' A stub implementation for the xbmcgui Dialog class textviewer() method '''
+        heading = kodi_to_ansi(heading)
+        text = kodi_to_ansi(text)
+        print('[37;100mTEXTVIEWER:[35;0m [%s] [35;0m%s[0m' % (heading, text))
```
dagwieers: Very nice. Since you know have an implementation of textviewer. You might as well call the about dialog in the test-framework, and get bonus points for coverage :-)

resources/settings.xml
```diff
@@ -1,11 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <settings>
-    <setting id="last_update" value="" visible="false" />
-    <setting id="version" value="" visible="false" />
+    <setting id="last_update" value="" visible="false"/>
+    <setting id="ish_version" value="" visible="false"/>
```
dagwieers: We need this in place as "version" as this is a standardized way to find if the addon has been upgraded and take action.

lib/inputstreamhelper.py
```diff
@@ -719,15 +726,15 @@ def _first_run():
         """Check if this add-on version is running for the first time"""
 
         # Get versions
-        settings_version = ADDON.getSetting('version')
+        settings_version = ADDON.getSetting('ish_version')
```
dagwieers: I would leave this as-is.

lib/inputstreamhelper.py
```diff
@@ -719,15 +726,15 @@ def _first_run():
...
         if settings_version == '':
             settings_version = '0.3.4'  # settings_version didn't exist in version 0.3.4 and older
         addon_version = ADDON.getAddonInfo('version')
 
         # Compare versions
         if LooseVersion(addon_version) > LooseVersion(settings_version):
             # New version found, save addon_version to settings
-            ADDON.setSetting('version', addon_version)
+            ADDON.setSetting('ish_version', addon_version)
```
dagwieers: I would leave this as-is.

lib/inputstreamhelper.py
```diff
@@ -206,7 +206,7 @@ def _arch(cls):
             if arm_version:
                 arch = 'armv' + arm_version.group()
         if arch in config.ARCH_MAP:
-            return config.ARCH_MAP[arch]
+            arch = config.ARCH_MAP[arch]
```
dagwieers: In this case returning early would be fine.
So I would leave it as it was.

lib/inputstreamhelper.py
```diff
@@ -668,14 +674,15 @@ def _install_widevine_arm(self):
                     self._set_loop_dev(),
                     self._losetup(bin_path),
                     self._mnt_loop_dev(),
-                ]
+                    ]
```
dagwieers: This indentation was fine.

test/xbmcgui.py
dagwieers: You can add it to `test/test_router.py` like the existing examples.

michaelarnauts:
What about using a shortened URL for this bug reports? This one is quite long to type over.

test/xbmcgui.py
horstle: Will do.

resources/settings.xml
horstle: Alright, I just thought since we have chromeos version now, too, it would be easier to distinguish.

lib/inputstreamhelper.py
horstle: will revert.

lib/inputstreamhelper.py
horstle: will revert.

lib/inputstreamhelper.py
horstle: This is a leftover from having "About" as settings page, where I didn't want to write `setSetting()` twice. Will revert.

lib/inputstreamhelper.py
horstle: Not sure how this got there...

horstle:
Good idea. I created https://git.io/fjp3K .

horstle:
No idea what that fail means.

dagwieers:
I am firmly against shortened URLs. At least the long URL has meaning, you know it is on GitHub, and that it is the official location of the source code. It is something a brain can remember by association.

horstle:
In general I don't really like shortened URLs either, because you can never know where it actually links to, but I think most users reporting bug would prefer to type a short meaningless URL rather than a long one. Especially since its easy to create a typo in "emilsvennesson" (no offense :)).

How about showing both? They might even fit in one line, like
> **https://git.io/fjp3K** (https://github.com/emilsvennesson/script.module.inputstreamhelper/issues)

dagwieers:
Yeah, I am also not a big fan of how codecov is now integrated. I guess the idea is that every new commit should not bring down coverage (by some margin). But that is not always possible or wanted. In the worst case scenario you may be spending more time on having something minor tested, than you spend on the original change. I don't know if we can modify the codecov integration though. I would also disable @codecov-io, it is a distraction IMO.

dagwieers:
The @codecov-io bot can be disabled on the codecov.io website, but I guess only the owner can do that. https://docs.codecov.io/v1.0/docs/codecov-yaml

The codecov.yml needs to contain:
```yaml
comment: off
```

horstle:
Alright, the 2 URLs don't fit into 1 line.

Bildschirmfoto von 2019-09-01 19-02-08
Bildschirmfoto von 2019-09-01 19-02-36

lib/inputstreamhelper.py
```diff
@@ -256,6 +256,14 @@ def _inputstream_version(self):
         addon = xbmcaddon.Addon(self.inputstream_addon)
         return addon.getAddonInfo('version')
 
+    @staticmethod
+    def _get_lib_version(lib):
+        if lib:
+            with open(lib, 'rb') as f:
+                x = re.search('[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+', str(f.read())) # pylint: disable=anomalous-backslash-in-string
```
mediaminister: When you define a regex pattern in Python, use a `r` before the quotation marks. Pylint won't complain anymore.
For instance:
```python
x = re.search(r'[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+', str(f.read()))
```

horstle: done.

mediaminister:
I removed the short url, so everything is displayed on one page. I fixed some linting warnings and I also rebased it to one commit so this can be cleanly merged.

For me, this is ready to be merged. @dagwieers @horstle

dagwieers:
Go for it !

dagwieers:
@horstle Thanks for spending the effort implementing this. It is very much appreciated. And it shipped today with v0.4.2 ! :smile:

horstle:
You already changed half of it again, but anyway, you're welcome. :)

dagwieers:
Not nearly, the changes I made were mostly superficial or visual.
(and from the perspective of add-on developers using the information window, like pietje666/plugin.video.vrt.nu#480)

Issue #132, opened 2019-08-30
ERROR: Unable to find plugin script.module.inputstreamhelper


horstle:
Hi, this issue can probably be closed right away, but I still thought I should report this:

I did a clean install of an addon relying on inputstreamhelper and updated the helper to the master branch right afterwards. Then I tried to run it via the settings page, but got the error in the title.
The addon I installed was able to run it and a restart of Kodi immediately fixed this, so I guess it's more of a Kodi issue and can be closed (here).

System is Kodi 18.3 from the official Repo on Ubuntu 18.04.3.

dagwieers:
I would like to get a reproducer of this to understand what is happening.

horstle:
Ok, so what I did was basically:

* On Ubuntu 18.04.3 install Kodi via `sudo apt install kodi`
* Install an addon with which inputstreamhelper will be installed, in my case Amazon VOD (this step might not be necessary)
* Download and install the zip of the master branch from Github
* Open the inputstreamhelper settings
* Try running either (re)install or remove. No matter which, you will see no reaction and the above error will be logged (and nothing more).

dagwieers:
@horstle It would be interesting to see if modifications to `addon.xml` would influence this behavior.

horstle:
The problem seems to be that script.module.routing does not install correctly. If I go to
settings > system > add-ons > manage dependencies > Plugin routing
and click on update, it works afterwards.

I did also try this by only installing the ZIP of 0.3.5 (instead of an addon relying on inputstreamhelper) and immediately afterwards the ZIP of master and could reproduce the same thing.
Only installing master without any previous install seems to correctly install script.module.routing.

horstle:
My guess is that Kodi might not reload PYTHONPATH in this situation (only after a manual update).

dagwieers:
I have seen that exact issue before, modules that have been installed just recently do not end up in PYTHONPATH and cause an import error.

horstle:
By the way, I was able to reproduce this in LibreELEC 9.1.002 on an RPi3, too.

mediaminister:
@horstle Can you confirm this is fixed now in the master branch?

horstle:
Yes, confirmed. Closing this.

PR #141, opened 2019-09-01
implement api using runscript

mediaminister:
This pull request includes:
- remove script.module.routing
- remove extension point xbmc.python.pluginsource
- use `RunScript` in settings.xml actions
- edit addon.py to use `RunScript` arguments


mediaminister:
@dagwieers @emilsvennesson @basrieter Please test and review, I think this can fix the blocked 0.4.0 release.

addon.py
```diff
@@ -3,46 +3,39 @@
...
+    if len(params) >= 2:
+        if params[1] == 'widevine_install':
+            widevine_install()
+        elif params[1] == 'widevine_remove':
+            widevine_remove()
+        elif params[1] == 'check_inputstream':
+            if len(params) == 3:
+                check_inputstream(params[2])
+            elif len(params) == 4:
+                check_inputstream(params[2], drm=params[3])
```
horstle: If there are more than 4 arguments specified or the first is neither of the 3, it would just do nothing. I think adding/logging some error like "invalid API call" would make sense.

basrieter:
The thing is that you can't pass params using RunScript() as far as I know. So it makes little sense to check them. I think each action will require its own script that can be called.

addon.py
mediaminister: Thanks for noticing, I edited my pull request accordingly.

basrieter:
I think this will do the trick?

horstle:
I tested this a bit on Ubuntu, seems to work well.

horstle:
It also works on LibreELEC on an RPi3. btw this would also fix #132 .

mediaminister:
Okay, tested and no problems found on Windows, Linux, LibreELEC and Android. I'm going to merge this.

Issue #146, opened 2019-09-02
unable update or re-install widevine for netflix


brunoduc:
install from [Odroid project](https://forum.odroid.com/viewtopic.php?f=98&t=34149)
Debug log :
```
kodi@omega:~# cat /home/kodi/.kodi/temp/kodi.log 
2019-09-02 09:46:53.798 T:2968072736  NOTICE: -----------------------------------------------------------------------
2019-09-02 09:46:53.798 T:2968072736  NOTICE: Starting Kodi (18.3 Git:20190619-89472b7d69-dirty). Platform: Linux ARM 32-bit
2019-09-02 09:46:53.798 T:2968072736  NOTICE: Using Release Kodi x32 build
2019-09-02 09:46:53.798 T:2968072736  NOTICE: Kodi compiled 2019-06-22 by GCC 7.4.0 for Linux ARM 32-bit version 4.15.18 (266002)
2019-09-02 09:46:53.799 T:2968072736  NOTICE: Running on Ubuntu 18.04.3 LTS, kernel: Linux ARM 32-bit version 4.14.133-166
2019-09-02 09:46:53.799 T:2968072736  NOTICE: FFmpeg version/source: 4.0.3-Kodi
2019-09-02 09:46:53.799 T:2968072736  NOTICE: Host CPU: ARMv7 Processor rev 3 (v7l), 8 cores available
2019-09-02 09:46:53.799 T:2968072736  NOTICE: ARM Features: Neon enabled
2019-09-02 09:46:53.799 T:2968072736  NOTICE: special://xbmc/ is mapped to: /usr/share/kodi
2019-09-02 09:46:53.799 T:2968072736  NOTICE: special://xbmcbin/ is mapped to: /usr/lib/arm-linux-gnueabihf/kodi
2019-09-02 09:46:53.799 T:2968072736  NOTICE: special://xbmcbinaddons/ is mapped to: /usr/lib/arm-linux-gnueabihf/kodi/addons
2019-09-02 09:46:53.799 T:2968072736  NOTICE: special://masterprofile/ is mapped to: /home/kodi/.kodi/userdata
2019-09-02 09:46:53.799 T:2968072736  NOTICE: special://envhome/ is mapped to: /home/kodi
2019-09-02 09:46:53.799 T:2968072736  NOTICE: special://home/ is mapped to: /home/kodi/.kodi
[...]

PR #148, opened 2019-09-02
make sure addon.xml meets all requirements

mediaminister:
This pull request includes the following changes to addon.xml:
- added standalone directive to xml declaration
- edit addon element attributes order
- added required summary element
- clean up spacing

### DIFF
--- a/addon.xml
+++ b/addon.xml
@@ -1,42 +1,44 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<addon id="script.module.inputstreamhelper" version="0.4.1" name="InputStream Helper" provider-name="emilsvennesson,dagwieers">
-    <requires>
-        <import addon="xbmc.python" version="2.25.0"/>
-        <!-- This is needed for the plugin to import the library -->
-        <import addon="script.module.inputstreamhelper" optional="true"/>
-        <import addon="script.module.kodi-six" version="0.0.4"/>
-        <import addon="script.module.pysocks" version="1.6.8" optional="true"/>
-    </requires>
-    <extension point="xbmc.python.module" library="lib"/>
-    <!-- This is needed to get an addon icon -->
-    <extension point="xbmc.python.script" library="addon.py"/>
-    <extension point="xbmc.addon.metadata">
-        <description lang="en_GB">A simple Kodi module that makes life easier for add-on developers relying on InputStream based add-ons and DRM playback.</description>
-        <description lang="el_GR">Μία απλή μονάδα για το Kodi η οποία διευκολύνει την ζωή των προγραμματιστών οι οποίοι εξαρτώνται από τα πρόσθετσ InputStream και αναπαραγωγή τύπου DRM.</description>
-        <description lang="ru_RU">Простой модуль для Kodi, который облегчает жизнь разработчикам дополнений, с использованием InputStream дополнений и воспроизведения DRM контента.</description>
-        <news>
-v0.4.1 (2019-09-01)
-- Use local url variable
-- Directly use Kodi CDM directory
-- Implement settings menu and API
-- Add integration tests
-- Add a progress dialog for extraction on ARM
-- Fix crash when using platform.system()
-- Fix a python error
-- Remove legacy Widevine CDM support
-- Replace requests/urllib3 with urllib/urllib2
-- Various unicode fixes
-- Add proxy support
-- Add setting to disable inputstreamhelper
-- Check Widevine support before all checks
-- Support 64-bit kernel with 32-bit userspace
-- Dutch, German, Greek, Italian, Russian and Swedish translations
-        </news>
-        <platform>all</platform>
-        <license>MIT</license>
-        <source>https://github.com/emilsvennesson/script.module.inputstreamhelper</source>
-        <assets>
-            <icon>resources/icon.png</icon>
-        </assets>
-    </extension>
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="script.module.inputstreamhelper" name="InputStream Helper" version="0.4.1" provider-name="emilsvennesson, dagwieers">
+  <requires>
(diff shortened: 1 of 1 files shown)


addon.xml
```diff
@@ -1,42 +1,43 @@
...
-        </news>
-        <platform>all</platform>
-        <license>MIT</license>
-        <source>https://github.com/emilsvennesson/script.module.inputstreamhelper</source>
-        <assets>
-            <icon>resources/icon.png</icon>
-        </assets>
-    </extension>
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="script.module.inputstreamhelper" name="InputStream Helper" version="0.4.1" provider-name="emilsvennesson,dagwieers">
```
dagwieers: So there needs to be a whitspace between two names.
I removed it myself by accident.

addon.xml
```diff
@@ -1,42 +1,43 @@
...
+  <extension point="xbmc.python.module" library="lib"/>
+  <!-- This is needed to get an addon icon -->
+  <extension point="xbmc.python.script" library="addon.py"/>
+  <extension point="xbmc.addon.metadata">
+    <summary lang="en_GB">Kodi InputStream and DRM playback made easy</summary>
+    <description lang="en_GB">A simple Kodi module that makes life easier for add-on developers relying on InputStream based add-ons and DRM playback.</description>
+    <description lang="el_GR">Μία απλή μονάδα για το Kodi η οποία διευκολύνει την ζωή των προγραμματιστών οι οποίοι εξαρτώνται από τα πρόσθετσ InputStream και αναπαραγωγή τύπου DRM.</description>
+    <description lang="ru_RU">Простой модуль для Kodi, который облегчает жизнь разработчикам дополнений, с использованием InputStream дополнений и воспроизведения DRM контента.</description>
+    <news>
+      v0.4.1 (2019-09-01)
```
dagwieers: There is no real need to indent this, and it's actually harder to manage if you have to copy the text over and also need to indent it.

addon.xml
```diff
@@ -1,42 +1,43 @@
...
-        <platform>all</platform>
-        <license>MIT</license>
-        <source>https://github.com/emilsvennesson/script.module.inputstreamhelper</source>
-        <assets>
-            <icon>resources/icon.png</icon>
-        </assets>
-    </extension>
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="script.module.inputstreamhelper" name="InputStream Helper" version="0.4.1" provider-name="emilsvennesson,dagwieers">
+  <requires>
```
dagwieers: Is there a reason why you move from 4 space indentation to 2 space indentation?

dagwieers:
You may want to add the About dialog from @horstle if you plan on releasing a new version.
I think it is already quite useful.

addon.xml
mediaminister: Yes, I strictly follow the recommendations at https://kodi.wiki/view/Addon.xml

dagwieers: Then we may want to do the same thing for the VRT NU add-on.

mediaminister:
Okay, I'll take a look at that, but there will be only a release if first https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/146 is fixed.

addon.xml
```diff
@@ -1,42 +1,43 @@
...
-        <assets>
-            <icon>resources/icon.png</icon>
-        </assets>
-    </extension>
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="script.module.inputstreamhelper" name="InputStream Helper" version="0.4.1" provider-name="emilsvennesson, dagwieers">
+  <requires>
+    <import addon="xbmc.python" version="2.25.0"/>
+    <!-- This is needed for the plugin to import the library -->
+    <import addon="script.module.inputstreamhelper" optional="true"/>
```
dagwieers: Is this still needed for a script?
Maybe the fact that we have a circular dependency is something Kodi can not cope with?

addon.xml
mediaminister: There is never a need for indenting xml, it's just cleaner.

addon.xml
mediaminister: Yes, last time I checked it was needed. I didn't change the logic of your code, I just removed the plugin interface.

addon.xml
```diff
@@ -1,42 +1,43 @@
...
+        - Various unicode fixes
+        - Add proxy support
+        - Add setting to disable inputstreamhelper
+        - Check Widevine support before all checks
+        - Support 64-bit kernel with 32-bit userspace
+        - Dutch, German, Greek, Italian, Russian and Swedish translations
+    </news>
+    <platform>all</platform>
+    <license>MIT</license>
+    <source>https://github.com/emilsvennesson/script.module.inputstreamhelper</source>
```
dagwieers: I still think it is silly we removed the &lt;website&gt; and &lt;forums&gt;, because pointing to the issues as forums, and the wiki as website is very convenient from the Kodi.tv website.

On the other hand, due to a bug (I think) the InputStream Helper is not listed on Kodi.tv, so we do not have a direct benefit at this time.

mediaminister: I just follow the rules. I don't think **script.modules** are listed on Kodi.tv.

> The <source> element provides the URL for the source code for this specific add-on. 

> The <website> element provides the website URL for this specific add-on. 

> The <forum> element provides the forum thread URL for this specific add-on. Leave this blank if there is no forum thread.
https://kodi.wiki/view/Addon.xml

dagwieers: At least a Wiki is also a website. (For VRT NU we used the Wiki)

The forum thread doesn't specify it is the Kodi forums thread (and it isn't checked either). So I wasn't convinced this had to be a Kodi forum thread specifically. (For VRTNU we used Facebook)

mediaminister: Okay, I readded the wiki website. But the issues listing is definitely not a forum.

Issue #147, opened 2019-09-02
Opening InputStream Helper info pane hangs Kodi


dagwieers:
When you open the InputStream Helper add-on information pane (either by clicking the add-on in the "Add-ons / Recently updated" menu, or by following the dependency of a package), your Kodi completely freezes.

This happened to me on LibreELEC v9.0.2 (Raspberry Pi) as well as on Windows 10 Kodi v18.4.

mediaminister:
This happens with all InputStream Helper versions on Kodi 18 and higher.
No problem on Kodi 17, the information pane just shows up.

mediaminister:
I found out there is a required element missing in addon.xml, I don't know if this is related to the freezing but it's definitely a good idea to fix this.  https://github.com/emilsvennesson/script.module.inputstreamhelper/pull/148

kekukui:
Confirmed on LibreELEC.  An update was offered from **0.3.5** to **0.4.1** but I cannot install it because the Kodi UI freezes when the property panel is opened.  On the RPi platform, CPU load on one core sticks at 100%.  (This is also a bug in Kodi because it should be possible to cancel instead of freezing the user interface.)  Same issue exists in 0.4.1.
______________________________

Kodi debug log
```
DEBUG: LIRC: - NEW 166 0 KEY_INFO devinput (KEY_INFO)
DEBUG: HandleKey: launch_media_center (0xc3) pressed, action is info
DEBUG: ------ Window Init (DialogAddonInfo.xml) ------
 INFO: Loading skin file: DialogAddonInfo.xml, load type: KEEP_IN_MEMORY
DEBUG: CAddonSettings[script.module.inputstreamhelper]: loading setting definitions
DEBUG: CAddonSettings[script.module.inputstreamhelper]: trying to load setting definitions from old format...
DEBUG: CAddonSettings[script.module.inputstreamhelper]: loading setting values

```

dagwieers:
@kekukui Please try release v0.4.0, I am confident it works.
https://github.com/emilsvennesson/script.module.inputstreamhelper/releases/download/v0.4.0/script.module.inputstreamhelper-0.4.0.zip

kekukui:
Same problem with 0.4.0.

mediaminister:
I did a couple of tests with a clean Windows 10 Kodi 18.4 install:
- rename existing add-on script.module.tzlocal to script.module.inputstreamhelper ==> Kodi freezes
[...]

PR #153, opened 2019-09-03
0.4.2 release

mediaminister:
Because 0.4.1 with https://github.com/emilsvennesson/script.module.inputstreamhelper/commit/0993eef6f5660d871559428cf89560bc9bff8f29 broke libwidevinecdm.so extraction from Chrome OS, I reverted this commit. Now we need to release a 0.4.2 version so InputStream Helper is functional again for people using platforms like Rpi and Odroid.
@basrieter @dagwieers @emilsvennesson Please test and approve

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,17 @@ Please report any issues or bug reports on the [GitHub Issues](https://github.co
 This module is licensed under the **The MIT License**. Please see the [LICENSE.txt](LICENSE.txt) file for details.
 
 ## Releases
+## v0.4.2 (2019-09-03)
+- Move release history to readme file (@mediaminister)
+- Clean up coverage/codecov config (@dagwieers)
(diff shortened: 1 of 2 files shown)


emilsvennesson:
We have to be more careful with changes like this in the future.

mediaminister:
Okay, I'm trying to fix https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/147 now.

dagwieers:
@emilsvennesson The rewrite of fdisk/parted was not intended to be released as it needed wider testing and unit tests. This was planned for the next release (v0.5.0).

The existing implementation has the assumption that the ChromeOS root filesystem is always the 3rd partition, which is a big risk for the future. The new implementation picks the biggest partition found, but in order to get to the biggest we need to parse every possible output of fdisk and parted, which, especially for fdisk, is a big issue.

For another project (Relax-and-Recover) we actually check the version in order to match the output.

Issue #94, opened 2019-08-24
Code structure


emilsvennesson:
The code in **inputstreamhelper.py** is starting to become rather complex and it's time we start splitting parts into different files. The new API @dagwieers implemented definietely belongs in its own file, and there are probably other parts that we could split up as well to make everything look a bit cleaner. I have something like this in mind:
* One file (**api.py**) for the API stuff.
* One file (**common.py**/**utils.py**) that includes all/most of the class methods and other code we can use across all files (JSON RPC requests, HTTP get functions etc).
* A folder (**widevine**) with three files: **arm.py**, **x86.py** and **widevine.py**. **widevine.py** would obviously include code that works on both platforms, and we could then import **arm.py**/**x86.py** depending on which platform the user is on. That way, we'll be able to simplify things and restructure the ugly **install_widevine()** wrapper method.
* **inputstreamhelper.py** includes code related to inputstream.adaptive such as enabling/disabling the add-on, HLS checks etc.

In the future, we should also split up the logging code and make use of the Python logging module. Another issue should be opened to discuss this though! :-)

-- Thoughts?

mediaminister:
I think we first need to increase the quality of the code in the current single file.
The code is too complex,  using too much global variables,  duplicating functionality over different functions, using a lot of external libraries while Kodi has a built-in Python API: http://romanvm.github.io/Kodistubs/, etc...

dagwieers:
And we need more/better tests as well. We don't have any unit tests at the moment.

dagwieers:
We cannot just move it to `api.py`. The problem is that currently everything is in `inputstreamhelper.py`, so we cannot change that interface. If we wanted an `api.py` it should go into `inputstreamhelper/api.py` but that means that `Helper()` will have to live in `inputstreamhelper/__init__py`.

So for the time being (in #98) I move the API to `inpustreamhelper_api.py`. In itself it doesn't matter where it lives, since nobody should be calling this directly (in contrast to `inpustreamhelper.py`.

We could also keep the API stuff in `addon.py` and don't bother with the recommendation to keep it under 15 lines. It doesn't matter to us anyway. I prefer this over `inputstreamhelper_api.py` as it won't be exposed to developers. (So I will revert it back ;-))

dagwieers:
[...]

PR #154, opened 2019-09-03
Restructure code for new release

### LINKED ISSUE #147: Opening InputStream Helper info pane hangs Kodi

dagwieers:
This restructures the code:
- Rename addon.py into ishelper.py
- Move api.py and config.py out of the global module namespace
- Fix tests to accommodate this

This fixes #147 
This relates to #94


dagwieers:
So while we did not want to reorganize the code already, the fact we are now exposing `api.py` in the module namespace would make think we have to do it now.

The main code is now in `__init__.py` for backward compatibility, and we probably want to split it up in the future too, but this requires a design decision that I don't want to have now. We probably want to think how we would like developers to use inputstreamhelper in the future.

dagwieers:
@emilsvennesson Did we disable codecov reporting completely at the org level?
I don't understand why there is no codecov check anymore.

emilsvennesson:
@dagwieers I haven't touched anything related to codecov. I guess we need to recheck the settings again.

mediaminister:
No problems found on Windows, Linux, LibreELEC, Android

default.py
```diff
@@ -2,7 +2,9 @@
 ''' This is the actual InputStream Helper API script entry point '''
 
 from __future__ import absolute_import, division, unicode_literals
+import os
 import sys
-from api import run
+sys.path.append(os.path.join(os.path.dirname(os.path.realpath(__file__)), 'lib'))  # noqa: E402
```
mediaminister: Will `os.path.realpath(__file__)` always work on all platforms? Maybe using the xbmc Python api for this is safer:
```python
from kodi_six import xbmc, xbmcaddon
sys.path.append(os.path.join(xbmc.translatePath(xbmcaddon.Addon().getAddonInfo('path')), 'lib'))
```

dagwieers: Yes, using os.path.realpath(__file__) is the best way to do it.

mediaminister: Okay, but I hope we don't  break the rules using `os.path.realpath(__file__)`
https://kodi.wiki/view/Add-on_rules#Requirements_for_scripts_and_plugins

dagwieers: I see other add-ons doing this too, but you can never be sure.

PR #155, opened 2019-09-03
Rename 'About' into 'IS Info'

dagwieers:
The reason to rename the 'About' dialog to 'InputStream Helper
information' is that it is clearer to the user where the information is
coming from exactly.

As we expect developers to use this from their own add-on for
troubleshooting purposes, it is important to make clear what it is about
(pun intended).

This PR includes:
- Renaming About dialog in Info dialog
- Dutch translations
- Localize '(disabled)' string
- Restructure translation strings (30800: information, 30900: settings)
- Prepare translation to other languages
- Small visual improvements

### DIFF
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -1014,33 +1014,30 @@ class Helper:
 
         return dict(http=proxy_address, https=proxy_address)
 
-    def about_dialog(self):
-        """ Show an About box with useful info e.g. for bug reports"""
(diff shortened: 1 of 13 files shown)


dagwieers:
@mediaminister Please take this on as well for v0.4.2 so we don't have to break the interface at a later stage.

mediaminister:
Looks good to me, just found something strange. Inside libwidevinecdm.so there is a `4.10.1440.18` string while the corresponding zip file states `4.10.1440.19`

dagwieers:
I had seen the same thing. I assume they shipped `4.10.1440.18` in a new wrapper.
As this is the case for all platforms/architectures.

mediaminister:
Okay, can this be merged?

PR #160, opened 2019-09-04
Update french translation with missing strings

dagwieers:
The contributed translation was based on the v0.4.1 release.


dagwieers:
@brunoduc I noticed the file you translated was based on the v0.4.1 release. I added the missing strings for the v0.4.2 release. If you would like to translate those as well, be my guest ;-)

Thanks again for your contribution !

PR #166, opened 2019-09-06
fix visibility in kodi 19

mediaminister:
In the alpha versions of Kodi 19 "Matrix" the InputStream Helper icon is not visible. This pull request fixes this.
Tested already on Linux, Windows and LibreELEC and this doesn't change behaviour on Kodi 18.x
Please test this on other platforms.


dagwieers:
Yeah, we remove the *executable* bit to avoid the double icon many GitHub moons ago ;-)

Issue #165, opened 2019-09-06
Widevine CDM version selection and rollback support


dagwieers:
This functionality has various aspects that are intertwined with:

- The ability to backup Widevine CDM versions (up to 5, configurable) when updating
- The ability to install/rollback a Widevine CDM version of your choice (from backup, or online)
- ~The ability to configure the interval to check for Widevine CDM updates~ #173
- The ability to pin to a specific version (so a way to disable updating Widevine)

We would also like to have a way to detect when Widevine playback fails (see #69), in that case pinning to a specific Widevine CDM would get overruled when it fails to work.

These features were listed in [the Ideas in our Wiki](https://github.com/emilsvennesson/script.module.inputstreamhelper/wiki/Ideas#settings)

A rationale for giving this kind of control to the user is available from issues like:
- [Incompatibility with newest Widevine on ARM systems (netflix#465)](https://github.com/asciidisco/plugin.video.netflix/issues/465)
- [Stuttering playback of streams using new widevine (inputstream.adaptive#305)](https://github.com/peak3d/inputstream.adaptive/issues/305)

Smeulf:
I really like the pin idea, and rollback!

Hope you can implement it soon :)

If I may ask for an extra option: give a way to save previous version on a configurable path, ie nfs or SMB, because it's a lot to keep on some Raspberry Pi (~2gb). That would be awesome.

Thanks!

dagwieers:
The implementation requires us (on ARM) to also keep track of the Chrome OS version in relation to the Widevine version. Since multiple Chrome OS releases ship with the same Widevine image (and we only know after extraction), recording which Widevine ships with what Widevine version will offer us a table to select the right Chrome OS image for a given Widevine release.

We may want to ship our own table as a reference, and InputStream Helper will continue to register newer versions going forward. I think this offers the best user experience on ARM.

Smeulf:
> From what I understood, not the full 2GB image is kept. The widevine files are extracted and the image is removed again.

So obvious and space saving!

the-isz:
Would be great to at least have an "Ignore" option when inputstreamhelper detects an outdated widevine version.

You probably already know, but just in case:

Right now, on ARM, you can't use plugins that use inputstreamhelper even though the outdated widevine version (4.10.1440.18) is able to play the streams. You only have the options "download an update that doesn't exist" and "don't even try to play the video".

I found that disabling inputstreamhelper is a workaround for the time being.

[...]

PR #173, opened 2019-09-16
Make update frequency configurable.

horstle:
Before really looking into #165 I thought I'd quickly add this. Tested it a bit, seems to work well.

### DIFF
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -759,7 +759,7 @@ class Helper:
         last_update = ADDON.getSetting('last_update')
         if last_update and not self._first_run():
             last_update_dt = datetime.fromtimestamp(float(ADDON.getSetting('last_update')))
-            if last_update_dt + timedelta(days=config.WIDEVINE_UPDATE_INTERVAL_DAYS) >= datetime.utcnow():
+            if last_update_dt + timedelta(days=int(ADDON.getSetting('update_frequency'))) >= datetime.utcnow():
                 log('Widevine update check was made on {date}', date=last_update_dt.isoformat())
                 return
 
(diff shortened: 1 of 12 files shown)


lib/inputstreamhelper/__init__.py
```diff
@@ -759,7 +759,7 @@ def _update_widevine(self):
         last_update = ADDON.getSetting('last_update')
         if last_update and not self._first_run():
             last_update_dt = datetime.fromtimestamp(float(ADDON.getSetting('last_update')))
-            if last_update_dt + timedelta(days=config.WIDEVINE_UPDATE_INTERVAL_DAYS) >= datetime.utcnow():
+            if last_update_dt + timedelta(days=int(ADDON.getSetting('freq'))) >= datetime.utcnow():
```
dagwieers: I would go for `update_frequency` here.

resources/settings.xml
```diff
@@ -10,5 +10,6 @@
         <setting label="30905" help="30906" type="action" id="install_widevine" action="RunScript(script.module.inputstreamhelper, widevine_install)" enable="eq(-2,false)" visible="!system.platform.android"/>
         <setting label="30907" help="30908" type="action" id="remove_widevine" action="RunScript(script.module.inputstreamhelper, widevine_remove)" enable="eq(-3,false)" visible="!system.platform.android"/>
         <setting label="30909" help="30910" type="folder" id="temp_path" source="" option="writeable" default="special://masterprofile/addon_data/script.module.inputstreamhelper"/>
+        <setting label="30911" help="30912" type="number" id="freq" default="14" />
```
dagwieers: So here as well `update_frequency`. Also remove that trailing whitespace in the XML statement while you're at it ;-)

test/userdata/addon_settings.json
```diff
@@ -7,6 +7,7 @@
         "disabled": "false",
         "last_update": "",
         "temp_dir": "special://masterprofile/addon_data/script.module.inputstreamhelper",
-        "version": ""
+        "version": "",
+        "freq": "14"
```
dagwieers: Also `update_frequency`.

Issue #89, opened 2019-08-21
Improving performance


dagwieers:
So now that we have unit tests, the logging output shows a few things we are doing on every run:
```log
Debug: [script.module.inputstreamhelper-0.3.5]: Platform information: ('Linux', 'moria.gent.internal', '5.2.6-1.el7.elrepo.x86_64', '#1 SMP Sun Aug 4 10:13:32 EDT 2019', 'x86_64', '')
Debug: [script.module.inputstreamhelper-0.3.5]: inputstreamhelper is enabled. You can disable inputstreamhelper by setting "disabled" to "true" in settings.xml         (Note: only recommended for developers knowing what they're doing!)
Debug: [script.module.inputstreamhelper-0.3.5]: jsonrpc payload: {u'params': {u'addonid': u'inputstream.adaptive'}, u'jsonrpc': u'2.0', u'id': 1, u'method': u'Addons.GetAddonDetails'}
Debug: [script.module.inputstreamhelper-0.3.5]: jsonrpc response: {"id":1,"jsonrpc":"2.0","result":{"addon":{"enabled": "true"}}}
Debug: [script.module.inputstreamhelper-0.3.5]: inputstream.adaptive is installed.
Debug: [script.module.inputstreamhelper-0.3.5]: jsonrpc payload: {u'params': {u'properties': [u'enabled'], u'addonid': u'inputstream.adaptive'}, u'jsonrpc': u'2.0', u'id': 1, u'method': u'Addons.GetAddonDetails'}
Debug: [script.module.inputstreamhelper-0.3.5]: jsonrpc response: {"id":1,"jsonrpc":"2.0","result":{"addon":{"enabled": "true"}}}
Debug: [script.module.inputstreamhelper-0.3.5]: inputstream.adaptive 0.3.5 is enabled.
Debug: [script.module.inputstreamhelper-0.3.5]: inputstream.adaptive 0.3.5 is installed and enabled.
Debug: [script.module.inputstreamhelper-0.3.5]: Found Widevine binary at /home/dag/home-made/script.module.inputstreamhelper/test/cdm/libwidevinecdm.so
Debug: [script.module.inputstreamhelper-0.3.5]: [u'ldd', u'/home/dag/home-made/script.module.inputstreamhelper/test/cdm/libwidevinecdm.so'] cmd executed successfully.
[...]

PR #174, opened 2019-09-16
Add our own kodiutils library

dagwieers:
This PR includes:
- Introduce a kodiutils library
- Import some library to blocks where they are needed
- Import only the stuff we need from libraries
- Add proxy-support testing
- Backport xbmc stub libraries

This relates to #94 and #89


dagwieers:
@mediaminister We test both with and without proxy. In fact all tests run without proxy, but I do an x64 test with proxy. You only need to install proxy.py (`pip install proxy.py` or `pip install -r requirements.txt`)

lib/inputstreamhelper/__init__.py
```diff
@@ -136,7 +93,7 @@ def _diskspace(cls):
     @classmethod
     def _temp_path(cls):
         ''' Return temporary path, usually ~/.kodi/userdata/addon_data/script.module.inputstreamhelper/temp '''
-        temp_path = xbmc.translatePath(os.path.join(ADDON.getSetting('temp_path'), 'temp'))
+        temp_path = xbmc.translatePath(os.path.join(get_setting('temp_path', 'special://masterprofile/addon_data/script.module.inputstreamhelper'), 'temp'))
```
mediaminister: I just noticed this bug while testing:
xbmc.translatePath doesn't support unicode in Python2, we need to add `to_unicode()` to support unicode paths in Python 2

```python
to_unicode(xbmc.translatePath(os.path.join(get_setting('temp_path', 'special://masterprofile/addon_data/script.module.inputstreamhelper'), 'temp')))
```

dagwieers: Maybe good to add a `translate_path()` function that covers this correctly.

dagwieers: I noticed most calls were already doing this.

dagwieers: Fixed.

mediaminister:
Looks good to me.

PR #175, opened 2019-09-17
create all folders in temp path

mediaminister:
When running tests, we should create the full path.


dagwieers:
That's fine for me too. I didn't mind how it was as long as we understood what was happening.
There was no conflict in files for testing.

Issue #131, opened 2019-08-30
InputStream Helper API methods crash if InputStream Adaptive is not installed


mediaminister:
It's possible to install InputStream Helper and call API methods when InputStream Adaptive add-on is not installed or disabled, this causes a crash.
Before executing an API method, we should check if InputStream Adaptive add-on is installed/enabled and try to enable or install it automatically.

Debug log:
```
2019-08-30 23:20:35.811 T:3976   DEBUG: [script.module.inputstreamhelper][routing] Dispatching to 'widevine_remove', args: {}
2019-08-30 23:20:35.812 T:3976   DEBUG: [script.module.inputstreamhelper-0.3.5]: Platform information: uname_result(system='Windows', node='DESKTOP-LSHI51P', release='10', version='10.0.17134', machine='AMD64', processor='Intel64 Family 6 Model 60 Stepping 3, GenuineIntel')
2019-08-30 23:20:35.816 T:3976   ERROR: EXCEPTION: Unknown addon id 'inputstream.adaptive'.
2019-08-30 23:20:35.878 T:3976   ERROR: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                             - NOTE: IGNORING THIS CAN LEAD TO MEMORY LEAKS!
                                            Error Type: <class 'RuntimeError'>
                                            Error Contents: Unknown addon id 'inputstream.adaptive'.
                                            Traceback (most recent call last):
                                              File "C:\Users\test\AppData\Roaming\Kodi\addons\script.module.inputstreamhelper\addon.py", line 48, in <module>
                                                plugin.run(sys.argv)
                                              File "C:\Users\test\AppData\Roaming\Kodi\addons\script.module.routing\lib\routing.py", line 115, in run
                                                self._dispatch(path)
                                              File "C:\Users\test\AppData\Roaming\Kodi\addons\script.module.routing\lib\routing.py", line 126, in _dispatch
                                                view_func(**kwargs)
                                              File "C:\Users\test\AppData\Roaming\Kodi\addons\script.module.inputstreamhelper\addon.py", line 44, in widevine_remove
                                                Helper('mpd', drm='widevine').remove_widevine()
                                              File "C:\Users\test\AppData\Roaming\Kodi\addons\script.module.inputstreamhelper\lib\inputstreamhelper.py", line 699, in remove_widevine
                                                widevinecdm = self._widevine_path()
[...]

PR #176, opened 2019-09-19
fix crashes when inputstream addon is missing

mediaminister:
This fixes https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/131

### DIFF
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -111,7 +111,10 @@ class Helper:
     @classmethod
     def _ia_cdm_path(cls):
         ''' Return the specified CDM path for inputstream.adaptive, usually ~/.kodi/cdm '''
-        addon = Addon('inputstream.adaptive')
+        try:
+            addon = Addon('inputstream.adaptive')
+        except RuntimeError:
+            return None
         cdm_path = translate_path(addon.getSetting('DECRYPTERPATH'))
         if not xbmcvfs.exists(cdm_path):
             xbmcvfs.mkdir(cdm_path)
@@ -138,9 +141,10 @@ class Helper:
         if widevine_cdm_filename is None:
             return False
 
-        widevine_path = os.path.join(cls._ia_cdm_path(), widevine_cdm_filename)
-        if xbmcvfs.exists(widevine_path):
-            return widevine_path
+        if cls._ia_cdm_path():
+            widevine_path = os.path.join(cls._ia_cdm_path(), widevine_cdm_filename)
+            if xbmcvfs.exists(widevine_path):
+                return widevine_path
 
         return False
 
@@ -223,7 +227,11 @@ class Helper:
 
     def _inputstream_version(self):
         ''' Return the requested inputstream version '''
-        addon = Addon(self.inputstream_addon)
+        try:
+            addon = Addon(self.inputstream_addon)
+        except RuntimeError:
+            return None
+
         return to_unicode(addon.getAddonInfo('version'))
 
     @staticmethod
@@ -413,7 +421,7 @@ class Helper:
     def _inputstream_enabled(self):
         """Returns whether selected InputStream add-on is enabled.."""
         data = execute_jsonrpc(dict(jsonrpc='2.0', id=1, method='Addons.GetAddonDetails', params=dict(addonid=self.inputstream_addon, properties=['enabled'])))
-        if data['result']['addon']['enabled']:
+        if data.get('result', {}).get('addon', {}).get('enabled'):
             log('{addon} {version} is enabled.', addon=self.inputstream_addon, version=self._inputstream_version())
             return True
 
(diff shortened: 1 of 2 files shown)


lib/inputstreamhelper/__init__.py
```diff
@@ -930,7 +937,7 @@ def check_inputstream(self):
 
     def info_dialog(self):
         """ Show an Info box with useful info e.g. for bug reports"""
-        disabled_str = ' ({disabled}'.format(disabled=localize(30054))
+        disabled_str = ' ({disabled})'.format(disabled=localize(30054))
```
dagwieers: Good catch, I didn't even notice that.

resources/settings.xml
```diff
@@ -7,8 +7,8 @@
         <setting label="30901" type="action" id="info" action="RunScript(script.module.inputstreamhelper, info)"/>
         <setting label="30903" help="30904" type="bool" id="disabled" default="false"/>
         <setting label="30904" type="text" enable="false"/> <!-- disabled_warning -->
-        <setting label="30905" help="30906" type="action" id="install_widevine" action="RunScript(script.module.inputstreamhelper, widevine_install)" enable="eq(-2,false)" visible="!system.platform.android"/>
-        <setting label="30907" help="30908" type="action" id="remove_widevine" action="RunScript(script.module.inputstreamhelper, widevine_remove)" enable="eq(-3,false)" visible="!system.platform.android"/>
+        <setting label="30905" help="30906" type="action" id="install_widevine" action="RunScript(script.module.inputstreamhelper, widevine_install)" enable="eq(-2,false)" visible="!system.platform.android + System.HasAddon(inputstream.adaptive)"/>
```
dagwieers: I would rather disable it if **inputstream.adaptive** is missing. So that it is clear that this functionality exists but is not accessible right now.

lib/inputstreamhelper/__init__.py
```diff
@@ -413,7 +420,7 @@ def _has_inputstream(self):
     def _inputstream_enabled(self):
         """Returns whether selected InputStream add-on is enabled.."""
         data = execute_jsonrpc(dict(jsonrpc='2.0', id=1, method='Addons.GetAddonDetails', params=dict(addonid=self.inputstream_addon, properties=['enabled'])))
-        if data['result']['addon']['enabled']:
+        if data.get('result') and data.get('result').get('addon').get('enabled'):
```
dagwieers: You could also do:
```python
        if data.get('result', {}).get('addon', {}).get('enabled')
```
It is safer in case the addon-key would be missing as well.

mediaminister: Okay, changed.

lib/inputstreamhelper/__init__.py
```diff
@@ -223,8 +227,11 @@ def enable():
 
     def _inputstream_version(self):
         ''' Return the requested inputstream version '''
-        addon = Addon(self.inputstream_addon)
-        return to_unicode(addon.getAddonInfo('version'))
+        try:
+            addon = Addon(self.inputstream_addon)
+            return to_unicode(addon.getAddonInfo('version'))
```
dagwieers: I would move the return out of the except. Even if we would never expect a RuntimeError in that line, it is always a good practice to keep the stuff in an except-block to the bare minimum as possible

dagwieers:
I guess we can prepare for a v0.4.3 release when this is merged.
Lots of translations updates, and some bug-fixes and new functionality.

We probably want to call out for translations again and integrate a language test (optionally) to see if any of the languages is lacking strings.

lib/inputstreamhelper/__init__.py
mediaminister: Edited.

resources/settings.xml
mediaminister: I can't get this to work with the `eq(-2,false)` comparator. Maybe I'm doing something wrong, maybe Kodi doesn't support this.

dagwieers: You are right, I cannot find any examples in existing add-ons that are doing this. I was expecting something like this to work, but you probably already tested this:

`enable="eq(-2,false) + System.HasAddon(inputstream.adaptive)"```

dagwieers: While testing, it seems I cannot make it reload my settings easily. So testing requires to reinstall the **inputstreamhelper** addon :-/

dagwieers: Despite this being pretty clear it should work: https://kodi.wiki/view/Add-on_settings#Conditions
Maybe this is related to the rewrite (and compatibility layer) not working correctly: https://forum.kodi.tv/showthread.php?tid=321370

dagwieers: Most likely you need to add an invisible entry and reference it using `eq(-X,bool)` as well.

mediaminister: No, I tried with the invisible entry, but I can't get the right value from `System.HasAddon(inputstream.adaptive)`

dagwieers: If not, I don't mind merging this as-is. We can fix this if we move to the settings v2 interface (when we stop supporting Kodi v17)

mediaminister: I changed it so settings are disabled when inputstream.adaptive is missing and settings get invisible when the "Disable InputStream Helper" switch is activated.

mediaminister:
Okay, I will prepare the v0.4.3 release.

PR #181, opened 2019-09-20
Rename "Troubleshooting" to "Expert"

dagwieers:
It seems most other add-ons use "Expert" which better fits this section
here as well.

Also, even (or especially) when InputStream Helper is disabled the user
may want to install or remove Widevine. So it does not make sense to
disable this functionality when InputStream Helper is disabled.

I also added separators since having more than one section seems worse
than this.

This PR includes:
- Reorganization of settings
- Rename "Troubleshooting" into "Expert"
- Turn frequency into slider
- Restore default frequency from "14" to "31"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,5 +70,5 @@ clean:
 	@echo -e "$(white)=$(blue) Cleaning up$(reset)"
 	find . test/ -name '*.pyc' -type f -delete
 	find . -name '__pycache__' -type d -delete
-	find test/userdata/ -mindepth 1 -type d -delete
+	find test/userdata/temp/ -mindepth 1 -type d -delete
 	rm -rf .pytest_cache/ .tox/ *.log
--- a/resources/language/resource.language.de_de/strings.po
+++ b/resources/language/resource.language.de_de/strings.po
@@ -278,8 +278,8 @@ msgstr ""
 
 ### SETTINGS
 msgctxt "#30900"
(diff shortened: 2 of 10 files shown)


dagwieers:
So this is a bit experimental. I would like to see what others think about this reorganization.

dagwieers:
The reason for increasing the default frequency back to 31 days are:

- We now have a way to force install Widevine, in times of need
- On ARM, the ChromeOS image version is updated more frequently
  - but does not necessarily have a newer Widevine (in fact it updates less frequently)
  - and this process is fairly painful (just when you would like to watch something)
- More frequently updating also means a higher risk of breaking a working situation

Preferably we would have a different frequency for Linux/ARM than other platforms, but that is not a possibility.

BTW Now that we get the version from the library, we should actually check the existing Widevine version to *versions.txt*, and not update if we have the latest.

mediaminister:
I'm okay with these changes. I think reorganising the widevine version check should be part of v0.5.0 because I'm not convinced the current `_get_lib_version` is failsafe. I don't understand why Google releases a .19 zip containing a .18 library version. This also happens on Linux and releasing functionality we don't understand is asking for problems.

dagwieers:
Agreed.

*BTW The .18 and .19 zip files contain the exact same file BTW, so it's not like they "forgot to update the version" or something like that.*

PR #182, opened 2019-09-20
Disable all options on Android

dagwieers:
Sadly I didn't think of this before and noticed this just yet on Android.

There is no need for having any of the inputstreamhelper options on Android.


mediaminister:
Looks good to me. But it seems like the visible attribute doesn't work for separators? Another Kodi settings bug?

dagwieers:
Yeah, that's something that should work IMO.
In our case the visual artefact is not a problem.
But I need to report this!

dagwieers:
I opened a ticket, but the issue (actually regression) was already known.
So I would leave it as-is, if this is fixed in Kodi it will work.

PR #184, opened 2019-09-20
clean userdata except settings

mediaminister:
Fixes an error when temp dir doesn't exist


dagwieers:
Ok, this makes it clear what the intention is.

PR #185, opened 2019-09-20
Backport xbmc stubs and avoid writing

dagwieers:
We now no longer write settings changes.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,5 +1,6 @@
 [MESSAGES CONTROL]
 disable=
+    bad-option-value,
     duplicate-code,
     line-too-long,
     old-style-class,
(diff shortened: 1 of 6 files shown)


dagwieers:
To avoid having to do this in the future I am considering making a separate repository with these xbmc stubs.

Potential problems:
- Customizations that are needed need to live somewhere, possibly we can override functionality or include a configuration file
- This will conflict with the already existing kodi-stubs project

At the very start we could just support our existing codebases in the code before making it more abstract. Als for testing the stubs we would need the projects that use it to run everything, which is also a bit backward ;-)

mediaminister:
> This will conflict with the already existing kodi-stubs project

Can this potential conflict be avoided?

In Kodi stubs documentation, I read this:
> You can use Kodistubs in combination with some mocking library, e.g. mock, to write unit tests for your addon code.

Any idea if there is an add-on that is doing this?

PR #193, opened 2019-09-22
backup and local rollback

horstle:
This adds the option to automatically create a backup of the installed lib (on update/reinstall or rollback) and rollback, keeping multiple backups according to the setting 'backups'.

This does not yet implement online rollback or pinning, but I think we should discuss (in #165) whether pinning is actually any different from disabling.

### DIFF
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -3,8 +3,8 @@
 from __future__ import absolute_import, division, unicode_literals
 import os
 from inputstreamhelper import config
-from .kodiutils import (browsesingle, execute_jsonrpc, get_addon_info, get_proxies, get_setting, kodi_to_ascii, localize, log,
-                        notification, ok_dialog, progress_dialog, set_setting, textviewer, translate_path, yesno_dialog)
+from .kodiutils import (browsesingle, execute_jsonrpc, get_addon_info, get_proxies, get_setting, get_userdata_path, kodi_to_ascii, localize,
+                        log, notification, ok_dialog, progress_dialog, select_dialog, set_setting, textviewer, translate_path, yesno_dialog)
 
 # NOTE: Work around issue caused by platform still using os.popen()
 #       This helps to survive 'IOError: [Errno 10] No child processes'
@@ -113,6 +113,17 @@ class Helper:
             mkdir(cdm_path)
 
         return cdm_path
+
+    @classmethod
+    def _backup_path(cls):
+        ''' Return the path to the cdm backups '''
+        import xbmcvfs
+
+        path = os.path.join(get_userdata_path(), 'backup')
+        if not xbmcvfs.exists(path):
+            xbmcvfs.mkdir(path)
+
+        return path
 
     @classmethod
     def _widevine_config_path(cls):
@@ -554,6 +565,51 @@ class Helper:
 
         return devices
 
+    def _remove_old_backups(self, backup_path):
+        """Removes old Widevine backups, if number of allowed backups is exceeded"""
+        from distutils.version import LooseVersion  # pylint: disable=import-error,no-name-in-module
+        import shutil
+
+        max_backups = int(get_setting('backups', '4'))
+        versions = sorted([LooseVersion(version) for version in os.listdir(backup_path)])
+
+        if len(versions) < 2:
+            return
+
+        if 'x86' in self._arch():
+            installed_version = self._load_widevine_config()['version']
+        else:
+            installed_version = self._select_best_chromeos_image(self._load_widevine_config())['version']
+
+        while len(versions) > max_backups + 1:
+            remove_version = str(versions[1] if versions[0] == installed_version else versions[0])
+            log('removing oldest backup which is not installed: {version}', version=remove_version)
+            shutil.rmtree(os.path.join(backup_path, remove_version))
+            versions = sorted([LooseVersion(version) for version in os.listdir(backup_path)])
+
+        return
+
+    def _install_cdm_from_backup(self, version):
+        """Copies files from specified backup version to cdm dir"""
+        import xbmcvfs
+
+        filenames = os.listdir(os.path.join(self._backup_path(), version))
+
+        for filename in filenames:
+            backup_fpath = os.path.join(self._backup_path(), version, filename)
+            install_fpath = os.path.join(self._ia_cdm_path(), filename)
+
+            if xbmcvfs.exists(install_fpath):
+                xbmcvfs.delete(install_fpath)
+
+            try:
+                os.link(backup_fpath, install_fpath)
+            except OSError:
+                xbmcvfs.copy(backup_fpath, install_fpath)
+
+        log('Installed CDM version {version} from backup', version=version)
+        self._remove_old_backups(self._backup_path())
+
     def _install_widevine_x86(self):
         """Install Widevine CDM on x86 based architectures."""
         cdm_version = self._latest_widevine_version()
@@ -565,8 +621,10 @@ class Helper:
         if downloaded:
             progress = progress_dialog()
             progress.create(heading=localize(30043), line1=localize(30044))  # Extracting Widevine CDM
+            self._unzip(os.path.join(self._backup_path(), cdm_version))
+
             progress.update(94, line1=localize(30049))  # Installing Widevine CDM
-            self._unzip(self._ia_cdm_path())
+            self._install_cdm_from_backup(cdm_version)
 
             progress.update(97, line1=localize(30050))  # Finishing
             self._cleanup()
@@ -574,9 +632,6 @@ class Helper:
                 return False
 
             if self._has_widevine():
-                if os.path.lexists(self._widevine_config_path()):
-                    os.remove(self._widevine_config_path())
-                os.rename(os.path.join(self._ia_cdm_path(), config.WIDEVINE_MANIFEST_FILE), self._widevine_config_path())
                 wv_check = self._check_widevine()
                 if wv_check:
                     progress.update(100, line1=localize(30051))  # Widevine CDM successfully installed.
@@ -651,16 +706,21 @@ class Helper:
                     self._mnt_loop_dev(),
                 ]
                 if all(success):
+                    import json
+
                     progress.update(91, line1=localize(30048))  # Extracting Widevine CDM
-                    self._extract_widevine_from_img()
+                    self._extract_widevine_from_img(os.path.join(self._backup_path(), arm_device['version']))
+                    with open(os.path.join(self._backup_path(), arm_device['version'], os.path.basename(config.CHROMEOS_RECOVERY_URL) + '.json'), 'w') as config_file:
+                        config_file.write(json.dumps(devices, indent=4))
+
                     progress.update(94, line1=localize(30049))  # Installing Widevine CDM
+                    self._install_cdm_from_backup(arm_device['version'])
+
                     progress.update(97, line1=localize(30050))  # Finishing
                     self._cleanup()
                     if self._has_widevine():
                         import json
                         set_setting('chromeos_version', arm_device['version'])
-                        with open(self._widevine_config_path(), 'w') as config_file:
-                            config_file.write(json.dumps(devices, indent=4))
                         wv_check = self._check_widevine()
                         if wv_check:
                             progress.update(100, line1=localize(30051))  # Widevine CDM successfully installed.
@@ -766,15 +826,19 @@ class Helper:
 
         return yesno_dialog(localize(30026), eula, nolabel=localize(30028), yeslabel=localize(30027))  # Widevine CDM EULA
 
-    def _extract_widevine_from_img(self):
+    def _extract_widevine_from_img(self, backup_path):
         ''' Extract the Widevine CDM binary from the mounted Chrome OS image '''
         import shutil
-        for root, dirs, files in os.walk(str(self._mnt_path())):  # pylint: disable=unused-variable
+        import xbmcvfs
+
+        for root, _, files in os.walk(str(self._mnt_path())):
             if str('libwidevinecdm.so') not in files:
                 continue
             cdm_path = os.path.join(root, 'libwidevinecdm.so')
             log('Found libwidevinecdm.so in {path}', path=cdm_path)
-            shutil.copyfile(cdm_path, os.path.join(self._ia_cdm_path(), 'libwidevinecdm.so'))
+            if not xbmcvfs.exists(backup_path):
+                xbmcvfs.mkdir(backup_path)
+            shutil.copyfile(cdm_path, os.path.join(backup_path, 'libwidevinecdm.so'))
             return True
 
         log('Failed to find Widevine CDM binary in Chrome OS image.')
@@ -845,7 +909,12 @@ class Helper:
 
     def _unzip(self, unzip_dir, file_to_unzip=None):
         ''' Unzip files to specified path '''
+        import xbmcvfs
+
         ret = False
+
+        if not xbmcvfs.exists(unzip_dir):
+            xbmcvfs.mkdirs(unzip_dir)
 
         import zipfile
         zip_obj = zipfile.ZipFile(self._download_path)
@@ -987,3 +1056,35 @@ class Helper:
 
         log('\n{info}'.format(info=kodi_to_ascii(text)), level=2)
         textviewer(localize(30901), text)
+
+    def rollback_libwv(self):
+        """Rollback lib to a version specified by the user"""
+        backup_path = self._backup_path()
+        versions = os.listdir(backup_path)
+
+        if 'x86' in self._arch():
+            show_versions = versions
+        else:
+            show_versions = []
+
+            for version in versions:
+                lib_version = self._get_lib_version(os.path.join(backup_path, version, config.WIDEVINE_CDM_FILENAME[system_os()]))
+                show_versions.append('{}    ({})'.format(lib_version, version))
+
+        if not show_versions:
+            notification(localize(30004), localize(30056))
+            return
+
+        if 'x86' in self._arch():
+            installed_version = self._load_widevine_config()['version']
+        else:
+            installed_version = self._select_best_chromeos_image(self._load_widevine_config())['version']
+        del show_versions[show_versions.index(installed_version)]
+
+        version = select_dialog(localize(30057), show_versions)
+        if version != -1:
+            log('Rollback to version {version}', version=versions[version])
+            self._install_cdm_from_backup(versions[version])
+            notification(localize(30037), localize(30051))  # Success! Widevine successfully installed.
+
+        return
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -229,6 +229,14 @@ msgctxt "#30055"
 msgid "There is not enough free disk space in the current temporary directory. Would you like to specify a different one to temporarily use for the Chrome OS Recovery Image (e.g. on a USB)?"
 msgstr ""
 
+msgctxt "#30056"
+msgid "No backups found!"
+msgstr ""
+
+msgctxt "#30057"
+msgid "Choose a backup to restore"
+msgstr ""
+
 
 ### INFORMATION DIALOG
 msgctxt "#30800"
@@ -307,4 +315,12 @@ msgstr ""
 
 msgctxt "#30911"
 msgid "Remove Widevine CDM library..."
+msgstr ""
+
+msgctxt "#30913"
+msgid "Number of backups"
+msgstr ""
+
+msgctxt "#30915"
+msgid "Rollback Widevine CDM library..."
 msgstr ""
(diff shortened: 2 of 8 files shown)


horstle:
I'll hopefully be able to fix those travis issues tomorrow.

dagwieers:
Correct, pinning really means not updating. The question is if we want to make a distinction between *everything* inputstreamhelper does (installing/enabling add-ons) or just updating.

horstle:
I fixed the pylint issues, but somehow travis doesn't work atm?

horstle:
That's the first time I contribute here, that travis actually found a bug and didn't just fail because of pylint. Anyway it's fixed now, so if you have time, please have a look.

lib/inputstreamhelper/__init__.py
```diff
@@ -546,6 +546,61 @@ def _chromeos_config(self):
 
         return devices
 
+    def _backup_libwv(self):
+        """Creates backup of installed lib, keeping multiple backups according to setting 'backups'"""
+        backups = int(get_setting('backups', '2'))
```
dagwieers: I would default to 4 or 5. Since we are talking about the library itself, the disk space required is quite limited.

lib/inputstreamhelper/__init__.py
```diff
@@ -546,6 +546,61 @@ def _chromeos_config(self):
...
+        bak_dir = os.path.join(self._ia_cdm_path(), 'backup')
```
dagwieers: I would go for backup_path :-)

horstle: In my opinion it should be dir, because it is a directory. path could be both, file or dir.

lib/inputstreamhelper/__init__.py
horstle: ok

horstle:
> I would save every file with the version attached to the filename. But keep it next to the other libraries in _cdm/_.
> 
> So I would expect to see something like this if the number of backups to keep is set to **5**:
> 
>     * libwidevinecdm-1.4.9.1088.so
> 
>     * libwidevinecdm-4.10.1146.0.so
> 
>     * libwidevinecdm-4.10.1192.0.so
> 
>     * libwidevinecdm-4.10.1196.0.so
> 
>     * libwidevinecdm-4.10.1440.18.so
> 
>     * libwidevinecdm.so
> 
> 
> We could make the last one a hardlink to the versioned library (so that removing one or the other does not impact).
> 
> Personally I would look inside the file so that it means the **.19** version will not be available. But if this would be there as well, I wouldn't mind that much. For ARM we need to look at the file anyway, so I think it is easier to get it from the file.

Hardlinks would probably need different approaches on Linux or Windows and the only advantage would be saving some space, but as you say that's not really a point.

I don't understand your point about the .19. I'm using `_get_lib_version()` so exactly what you want...

Getting a list with available versions would be more difficult if the versions where only available from the filenames, so I think the dict is the best solution.

lib/inputstreamhelper/__init__.py
horstle: I changed bak to backup, but kept the dir.

lib/inputstreamhelper/__init__.py
horstle: changed it to 4.

dagwieers:
> Hardlinks would probably need different approaches on Linux or Windows and the only advantage would be saving some space, but as you say that's not really a point.

Hardlinks (**os.link()**) work fine on Windows, but for Python 2 we probably want to fall back to **shutil.copyfile()** instead.

> I don't understand your point about the .19. I'm using _get_lib_version() so exactly what you want...

Sure, the only problem is that it may lead to confusion because the ZIP file is .19. So we need to be careful here. It is actually the same issue we have with the ChromeOS version and the widevine version. We need to keep track of the "container" version and the "widevine" version and make a conversion table we can use to check if we have the *latest* version. For non-ARM it will often be the same, but we know we can't trust this.

lib/inputstreamhelper/__init__.py
```diff
@@ -546,6 +546,61 @@ def _chromeos_config(self):
 
         return devices
 
+    def _backup_libwv(self):
+        """Creates backup of installed lib, keeping multiple backups according to setting 'backups'"""
+        backups = int(get_setting('backups', '4'))
+        backup_dir = os.path.join(self._ia_cdm_path(), 'backup')
```
mediaminister: Not sure it's a good idea to make backups to the InputStream Adaptive CDM directory because InputStream Adaptive add-on uses this directory in a way we can't control.
I think it's better to save all backups in our own userdata directory and only copy a single CDM when needed.

dagwieers: @mediaminister It is part of the library search path, but based on name AFAIK. So that shouldn't be a problem if we stick to a proper naming plan (not something that could accidentally be picked up). In fact on Linux it is common to have multiple versions next to each other (and a symlink for an unversioned name). Although I wouldn't use the Unix convention to avoid it being picked up accidentally ;-)

dagwieers: That said, I wouldn't mind if we decide to move it elsewhere. I just think it is nicer if it stays together. (and the hardlink shows which one is being used)

horstle: > Not sure it's a good idea to make backups to the InputStream Adaptive CDM directory because InputStream Adaptive add-on uses this directory in a way we can't control.
> I think it's better to save all backups in our own userdata directory and only copy a single CDM when needed.

So keeping the license file and widevine_config.json in the cdm dir is fine, but a backup dir isn't? You're the one who put these files there in #95 @mediaminister .

BTW I decided to put the backups in an extra dir to avoid confusing potential fuzzy searches or similar, I think it should be rather safe there.

horstle:
> Hardlinks (**os.link()**) work fine on Windows, but for Python 2 we probably want to fall back to **shutil.copyfile()** instead.

I'm still not in favor of hardlinks, until we can guarantee every system is using Python3 (or not support Windows...). That would only make things more complicated and thus prone to errors. Again, we wouldn't save much space by that and to know which version is installed use get_lib_version() or look at the widevine_config.json. Before #95 the lib was linked on Unix systems, but copied on Windows. I didn't see any reason for these 2 approaches then and still don't see any now.

> Sure, the only problem is that it may lead to confusion because the ZIP file is .19. So we need to be careful here. It is actually the same issue we have with the ChromeOS version and the widevine version. We need to keep track of the "container" version and the "widevine" version and make a conversion table we can use to check if we have the _latest_ version. For non-ARM it will often be the same, but we know we can't trust this.

It backs up the widevine_config.json adding the same extension. That should be enough in my opinion. (E.g. the .19 is called that there, too.)

lib/inputstreamhelper/__init__.py
mediaminister: Mozilla Firefox also stores the `LICENSE.txt` and `manifest.json` in the same directory as `libwidevinecdm.so`...

I'm not sure if it's fine or even useful that `manifest.json` is renamed to `widevine_config.json`, but maybe @emilsvennesson remembers why this is done: https://github.com/emilsvennesson/script.module.inputstreamhelper/commit/2641cadfcb3aad741138f827f0b59bfc4bf2e2a0

There is also a [Kodi add-on rule](https://kodi.wiki/view/Add-on_rules) that states the following:
> add-ons should store all their data in their own subfolder inside the addon_data directory. Access (read/write/delete) to any other files / folders is not allowed by default.

If a Kodi user decides to remove the InputStreamHelper add-on, the backup dir will exist forever...

Those are the reasons why I have my doubts.

horstle: Alright, that clears it up. I'll move the backup dir.

horstle:
The 'container' version is now in backed_versions.json, but never used. Should it be shown in the version list during rollback together with the version from `get_lib_version()`?

horstle:
Ok, I rewrote this. Now it downloads to a backup directory inside another directory named after the "container" version and then copies the files over to the cdm directory. It can easily be changed to linking, but I would prefer not doing so until everyone is using python3.

resources/language/resource.language.en_gb/strings.po
```diff
@@ -308,3 +316,11 @@ msgstr ""
 msgctxt "#30911"
 msgid "Remove Widevine CDM library..."
 msgstr ""
+
+msgctxt "#30913"
+msgid "Number of Backups"
```
mediaminister: Backup should be lowercase in English.

lib/inputstreamhelper/__init__.py
```diff
@@ -959,3 +996,19 @@ def info_dialog(self):
...
         Dialog().textviewer(localize(30901), text)
+
+    def rollback_libwv(self):
+        """Backup current lib and rollback to a version specified by the user"""
+        backup_path = self._backup_path()
+        versions = os.listdir(backup_path)
+        show_versions = []
+
+        for version in versions:
+            lib_version = self._get_lib_version(os.path.join(backup_path, version, config.WIDEVINE_CDM_FILENAME[system_os()]))
```
mediaminister: Shouldn't we rely completely on container version and only read the manifest.json?

lib/inputstreamhelper/__init__.py
```diff
@@ -744,14 +776,16 @@ def _widevine_eula(self):
...
+        ''' Extract the Widevine CDM binary from the mounted Chrome OS image to backup dir '''
+        for root, _, files in os.walk(str(self._mnt_path())):
             if str('libwidevinecdm.so') not in files:
                 continue
             cdm_path = os.path.join(root, 'libwidevinecdm.so')
             log('Found libwidevinecdm.so in {path}', path=cdm_path)
-            shutil.copyfile(cdm_path, os.path.join(self._ia_cdm_path(), 'libwidevinecdm.so'))
+            if not xbmcvfs.exists(backup_path):
+                xbmcvfs.mkdir(backup_path)
+            shutil.copyfile(cdm_path, os.path.join(backup_path, 'libwidevinecdm.so'))
```
mediaminister: Does this work on Windows and Mac?

resources/language/resource.language.en_gb/strings.po
horstle: Sorry, will change.

lib/inputstreamhelper/__init__.py
horstle: `_extract_widevine_from_img()` will only ever be called on ARM devices with Linux. If you compare the deleted line above, this has obviously worked before.

lib/inputstreamhelper/__init__.py
horstle: On ARM this "container version" would be the version of the ChromeOS image and we don't have a manifest.json there. I think it makes sense there at least. If you want to I could remove this on x86.

lib/inputstreamhelper/__init__.py
mediaminister: Okay, I see.

lib/inputstreamhelper/__init__.py
mediaminister: I think it's better to remove this for x86 because the versions in https://dl.google.com/widevine-cdm/versions.txt only match with the manifest.json
I tested and 4.10.1440.19 gets never backed up because the lib version matches 4.10.1440.18

horstle: Alright, will change it in a minute (though I always prefer to keep it platform independent).
I don't understand why 4.10.1440.19 would not be backed up, since `_get_lib_version()` is never called during the backup process. During my tests it worked.

resources/language/resource.language.en_gb/strings.po
horstle: changed.

lib/inputstreamhelper/__init__.py
```diff
@@ -959,3 +996,23 @@ def info_dialog(self):
...
+        if 'x86' in self._arch():
+            show_versions = versions
+        else:
+            show_versions = []
+
+            for version in versions:
+                lib_version = self._get_lib_version(os.path.join(backup_path, version, config.WIDEVINE_CDM_FILENAME[system_os()]))
+                show_versions.append('{}    ({})'.format(lib_version, version))
+
+        version = Dialog().select(localize(30057), show_versions)
```
mediaminister: Showing an empty `Dialog.select()` when the backup dir is empty is not nice. I think it's better to show a `Dialog().notification` with an error message to the user.

lib/inputstreamhelper/__init__.py
```diff
@@ -959,3 +996,23 @@ def info_dialog(self):
         text += localize(30830, url=config.ISSUE_URL)  # Report issues
 
         Dialog().textviewer(localize(30901), text)
+
+    def rollback_libwv(self):
+        """Backup current lib and rollback to a version specified by the user"""
+        backup_path = self._backup_path()
+        versions = os.listdir(backup_path)
```
mediaminister: Not sure if we should remove the current installed version from the version to rollback to. Is there a use case to rollback to the current version?

lib/inputstreamhelper/__init__.py
horstle: Rollback would probably only ever be used if it was installed at least once, so it would at least show the current version, but just to be safe I added a check.

lib/inputstreamhelper/__init__.py
horstle: Not really. As long as the backups are copied to the cdm dir and not linked, it might be useful if the installed version was somehow corrupted. When we switch to links, the only reason I can think of is when you decrease the number of backups this would call `remove_old_backups()` quicker than doing a reinstallation.
This reminds me that I forgot to add a check to `remove_old_backups()`, so it wouldn't remove the installed version. This wouldn't necessarily be a problem at the moment, but definitely when we change to links.

horstle: I added the check to `remove_old_backups()` and I removed the installed version.

mediaminister:
Looks good to me.
@dagwieers If you have no more remarks, I will merge this.

lib/inputstreamhelper/__init__.py
```diff
@@ -540,6 +549,37 @@ def _chromeos_config(self):
 
         return devices
 
+    def _remove_old_backups(self, backup_path):
+        """Removes old Widevine backups, if number of allowed backups is exceeded"""
+        backups = int(get_setting('backups', '4'))
```
dagwieers: I would rename this: `max_backups`

lib/inputstreamhelper/__init__.py
```diff
@@ -540,6 +549,37 @@ def _chromeos_config(self):
...
+        versions = sorted([LooseVersion(version) for version in os.listdir(backup_path)])
+
+        if 'x86' in self._arch():
+            installed_version = self._load_widevine_config()['version']
+        else:
+            installed_version = self._select_best_chromeos_image(self._load_widevine_config())['version']
+
+        while len(versions) > backups + 1:
```
dagwieers: I think this can be more simple. It should return when `backups == 0` or when `len(versions) <= 1`.
And `remove_version = str(versions[1] if remove_version == installed_version else versions[0])`
So you can get rid of the exception-handling.

dagwieers: So basically at the top:
```python
        if backups == 0 or len(versions) <= 1:
            return
```

And then here:
```python
        remove_version = str(versions[1] if remove_version == installed_version else versions[0])
```

lib/inputstreamhelper/__init__.py
```diff
@@ -540,6 +549,37 @@ def _chromeos_config(self):
...
+            log('removing oldest backup which is not installed: {version}', version=remove_version)
+            shutil.rmtree(os.path.join(backup_path, remove_version))
+            versions = sorted([LooseVersion(version) for version in os.listdir(backup_path)])
+
+    def _install_cdm_from_backup(self, version):
+        """Copies files from specified backup version to cdm dir"""
+        filenames = os.listdir(os.path.join(self._backup_path(), version))
+
+        for filename in filenames:
+            xbmcvfs.copy(os.path.join(self._backup_path(), version, filename), os.path.join(self._ia_cdm_path(), filename))
```
dagwieers: As I said multiple times now, we ought to use hard links where they are supported. And fall back to copying when it is not. Otherwise we may be breaking a working setup because the disk is full in cases where the fix already works fine.

We don't fix bugs on a basis of lowest common denominator.

lib/inputstreamhelper/__init__.py
```diff
@@ -540,6 +549,37 @@ def _chromeos_config(self):
...
+        log('Installed CDM version {version} from backup', version=version)
+        self._remove_old_backups(self._backup_path())
```
dagwieers: I know that it makes it a little bit more complex, but I would remove first before adding the new. This will also lower the risk of a disk full condition.

resources/settings.xml
```diff
@@ -10,8 +10,10 @@
         <setting label="30904" type="text" id="warning" enable="false"/> <!-- disabled_warning -->
         <setting label="30905" help="30906" type="slider" id="update_frequency" default="31" range="1,3,90" option="int" enable="eq(-2,false)" visible="!system.platform.android"/>
         <setting label="30907" help="30908" type="folder" id="temp_path" source="" option="writeable" default="special://masterprofile/addon_data/script.module.inputstreamhelper" visible="!system.platform.android"/>
+        <setting label="30913" help="30914" type="slider" id="backups" default="4" range="0,1,5" option="int" visible="!system.platform.android"/>
```
dagwieers: I would not limit this to 5. I would say, make it go up to 15 or 20.
This would be something useful for us in development too.

lib/inputstreamhelper/__init__.py
horstle: Done.

lib/inputstreamhelper/__init__.py
horstle: If the number of allowed backups is decreased, it has to remove multiple backups, so the while loop is necessary. I added a check at the top for `len(versions)` and simplyfied `remove_version`.

lib/inputstreamhelper/__init__.py
horstle: If it did that with allowed backups set to 0 and something went wrong, we'd have a situation like in #218, especially when using links. I thought I was supposed to fix that?

resources/settings.xml
horstle: #165, second line...
But since I thought these were a bit few, too, I'll increase it.

lib/inputstreamhelper/__init__.py
horstle: I really don't understand your obsession on those links (which are not used anywhere at the moment, so I'm not sure which fix you are talkling about). If your system ever has less than 10MB space left, you have more serious problems. I just wanted to keep the OS specific part to a minimum, otherwise we could just write a different addon for each system.

That being said, I'll add those links, since this is only a small change and not worth the whole discussion.

lib/inputstreamhelper/__init__.py
dagwieers: Correct, that's an oversight of me.

resources/settings.xml
dagwieers: @horstle Sorry for changing my mind during review. Back then I was not thinking of a slider with a limitation. (I do prefer the slider though)

lib/inputstreamhelper/__init__.py
dagwieers: No, because we wouldn't be removing the library we are using. But you are right that it wouldn't work as people expect when backups is set to 0.

Personally I would always leave one backup (the current library) behind so keep the lower limit to 1.

lib/inputstreamhelper/__init__.py
dagwieers: It is not an obsession, it's avoiding potential known problems. Why not do that when it works on most platforms. That said, if you don't want to do it, I don't mind adding it later.

lib/inputstreamhelper/__init__.py
```diff
@@ -540,6 +551,45 @@ def _chromeos_config(self):
...
+        for filename in filenames:
+            backup_fpath = os.path.join(self._backup_path(), version, filename)
+            install_fpath = os.path.join(self._ia_cdm_path(), filename)
+
+            if system_os() == 'Windows' and version_info[0] < 3:
+                xbmcvfs.copy(backup_fpath, install_fpath)
+            else:
+                if xbmcvfs.exists(install_fpath):
+                    xbmcvfs.delete(install_fpath)
+                os.link(backup_fpath, install_fpath)
```
dagwieers: Two comments:
- `os.link()` shouldn't require a delete, else it would not be atomic
- I think you can simply try `os.link()` and when you get an OSError exception perform a copy

This should work:
```python
            try:
                os.link(backup_fpath, install_fpath)
            except OSError:
                xbmcvfs.copy(backup_fpath, install_fpath)
```

dagwieers:
We probably should have merged this before the other PRs, then I would have resolved the conflicts.

@horstle If you prefer, I can resolve the conflicts for you in this PR.

lib/inputstreamhelper/__init__.py
horstle: `os.link()` gives an OSError on Python2 and a FileExistsError on python3 if the destination exists. I had checked that. But I'll change it to try...except.

dagwieers: This is truly unfortunate that anno 2019 there is no race-free implementation of `os.link()`.
https://bugs.python.org/issue36656

Revert to what you have, I'll crawl under a stone.

horstle:
It's fine, I resolved them.

dagwieers:
Let's get this merged, it will be easier for testing with the other PRs.
At some point in the future I would like to add integration tests to ensure proper behaviour.

horstle:
There were more conflicts than I thought, but they should be resolved now.

dagwieers:
Thanks @horstle for 6 weeks plus 1 year of patience and persisting despite pedantic reviews.

PR #197, opened 2019-09-25
fix new pylint v2.4 tests

mediaminister:
Importing modules at function level is considered a bad practice in pylint 2.4, this pull request moves all imports to module level.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -2,6 +2,7 @@
 disable=
     bad-option-value,
     duplicate-code,
+    import-outside-toplevel,
     line-too-long,
     old-style-class,
     too-many-branches,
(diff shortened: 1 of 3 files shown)


dagwieers:
IMO that is the wrong thing to do. Pylint implements general rules for programming without taking into account specifics. So it it up to us to evaluate what rules make sense and what not.

In our case for Kodi add-ons importing when needed makes the add-on feel more reactive because you spread the imports over different actions (they are still being cached).

The solution here is to disable this check IMO.

dagwieers:
BTW I wanted to fix this last night, but fell asleep when putting my son to bed :roll_eyes:

mediaminister:
My fault, I forgot about the problems we had in the past with the slow requests module and why we decided to use function level imports.
I totally agree speed is more important than PEP8 for add-ons that run on RPi's.

dagwieers:
Indeed, this is also more important on RPi than it is on other platforms.

PR #198, opened 2019-09-25
Move imports to where they are being used

dagwieers:
May be controversial, but if we want to keep inputstreamhelper's
footprint as small as possible to add-ons using it, this may help.

To be discussed.

### DIFF
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -1,29 +1,10 @@
 # -*- coding: utf-8 -*-
 ''' Implements the main InputStream Helper class '''
 from __future__ import absolute_import, division, unicode_literals
-
 import os
-import platform
-import json
-import subprocess
-import shutil
-from distutils.version import LooseVersion  # pylint: disable=import-error,no-name-in-module
-from datetime import datetime, timedelta
-
-try:  # Python 3
-    from urllib.error import HTTPError
-    from urllib.request import build_opener, install_opener, ProxyHandler, urlopen
-except ImportError:  # Python 2
-    from urllib2 import build_opener, HTTPError, install_opener, ProxyHandler, urlopen
-
 from inputstreamhelper import config
-
-import xbmc
-from xbmcaddon import Addon
-import xbmcvfs
 from .kodiutils import (browsesingle, execute_jsonrpc, get_addon_info, get_proxies, get_setting, kodi_to_ascii, localize, log,
                         notification, ok_dialog, progress_dialog, set_setting, textviewer, translate_path, yesno_dialog)
-from .unicodehelper import to_unicode
 
 # NOTE: Work around issue caused by platform still using os.popen()
 #       This helps to survive 'IOError: [Errno 10] No child processes'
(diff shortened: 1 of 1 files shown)


dagwieers:
My main objection to this PR would be that restructuring the code would make things a lot easier and could provide the same benefits.

Some of the things we move out of the global scope are actually things we need all over the place, however that may be true for inputstreamhelper, but not necessarily for the add-on importing inputstreamhelper.

We should probably advise add-on developers to only import inputstreamhelper when they need it, rather than globally as well.

dagwieers:
A fix for xbmcgui is in #202

michaelarnauts:
I'm doing this myself sometimes, but has somebody actually benchmarked this if this really is faster?

dagwieers:
@michaelarnauts It makes a difference on the very first startup (when `reuselanguageinvoker` is used) and on every interaction (when `reuselanguageinvoker` is not used). When importing urllib3 (when it took 4 seconds) this difference was very apparent and that is when we started using `reuselanguageinvoker`. On Raspberry Pi this effect is very real.

The VRT NU startup time is a few 100ms, whereas typical add-ons take a second or more. You can *feel* the difference, and see it from the default spinner. The difference between VRT NU and VTM GO is still there. With the newer upcoming urllib3 this difference will become smaller.

For a utility like **inputstreamhelper** we shouldn't be adding anything to the start-up time if that is not essential for its use IMO. That is why we dropped urllib3 (some add-ons avoid urllib3 anyway) and why we avoid importing things that are not used. (e.g. kodi-six)

And then we also have #89 to see how we can reduce the overhead of using **inputstreamhelper** when it *is* being used.

dagwieers:
Benchmarks on Raspberry Pi 3 are available from:
- https://github.com/pietje666/plugin.video.vrt.nu/wiki/Philosophy#importing-libraries-on-raspberry-pi

The upcoming urllib3 release will bring the current import times of urllib3 from 800ms (v1.25.3) to 350ms (v1.25.6) on Raspberry Pi.

dagwieers:
Also take note that in Kodi Matrix (Python 3) `reuselanguageinvoker` has been removed. So the startup time will be the interaction time. https://github.com/xbmc/xbmc/pull/16116#issuecomment-542222897

michaelarnauts:
> Also take note that in Kodi Matrix (Python 3) `reuselanguageinvoker` has been removed. So the current startup time will become the interaction time. [xbmc/xbmc#16116 (comment)](https://github.com/xbmc/xbmc/pull/16116#issuecomment-542222897)

This is terrible news, but by looking at your linked issue, it was not removed on purpose, so it might come back?

dagwieers:
@michaelarnauts Hmm, I missed that. I get the mails from that thread but didn't link the responses to  reuselanguageinvoker`. That is good news!

dagwieers:
BTW I also updated the Wiki to advise developers to import inputstreamhelper locally and not globally.

https://github.com/emilsvennesson/script.module.inputstreamhelper/wiki/Integration

dagwieers:
Ready to be merged.

PR #202, opened 2019-09-25
Autoinstall Widevine updates

dagwieers:
This adds a feature to enable automatic installation of Widevine.


dagwieers:
So, we may want to allow for this only after accepting the EULA once, or maybe as long as the EULA does not change. The idea is to have this when users are not technical savvy and may not continue as expected.

This would fix the situations I have seen where people (or children) did not proceed because they didn't know what to do (and the buttons default to No/Cancel).

resources/settings.xml
```diff
@@ -10,6 +10,7 @@
         <setting label="30904" type="text" id="warning" enable="false"/> <!-- disabled_warning -->
         <setting label="30905" help="30906" type="slider" id="update_frequency" default="31" range="1,3,90" option="int" enable="eq(-2,false)" visible="!system.platform.android"/>
         <setting label="30907" help="30908" type="folder" id="temp_path" source="" option="writeable" default="special://masterprofile/addon_data/script.module.inputstreamhelper" visible="!system.platform.android"/>
+        <setting label="Auto-install Widevine CDM" type="bool" id="automatic_install" default="false"/>
```
mediaminister: I think "Unattended Widevine CDM install" is a better description.

mediaminister:
> So, we may want to allow for this only after accepting the EULA once, or maybe as long as the EULA does not change.

I didn't see a Widevine CDM EULA when testing this. This can only be merged if the user accepts the EULA **every time** a user enables this setting.

resources/settings.xml
dagwieers: Agreed.

dagwieers:
@mediaminister Correct, I don't think this would be ready to be merged. I wasn't sure if we wanted this really, so I put it out here to discuss.

dagwieers:
Let me also move the non-related changes out of this PR.

Issue #206, opened 2019-10-05
Nvidia Shield widevine version not found


herbalizer404:
Hi,
I have shield experience 8.0.1
ish 0.4.3
Netflix plugin 0.15.3
But ish don't find widevine, have you an idea?
Thanks

dagwieers:
Is anything written in the kodi.log?
Do you get an error message?

herbalizer404:
2019-10-05 00:08:55.328 T:12752   ERROR: CDVDVideoCodecAndroidMediaCodec::AddData error(-10000)
2019-10-05 00:08:55.332 T:12756   ERROR: Previous line repeats 1 times.
2019-10-05 00:08:55.332 T:12756  NOTICE: Creating audio stream (codec id: 86056, channels: 6, sample rate: 48000, pass-through)
2019-10-05 00:08:55.343 T:12752   ERROR: CDVDVideoCodecAndroidMediaCodec::AddData error(-10000)
2019-10-05 00:08:55.453 T:12654   ERROR: Previous line repeats 8 times.
2019-10-05 00:08:55.453 T:12654  NOTICE: Trying to open: samplerate: 192000, channelMask: 12, encoding: 13
2019-10-05 00:08:55.462 T:12752   ERROR: CDVDVideoCodecAndroidMediaCodec::AddData error(-10000)
2019-10-05 00:08:55.468 T:12654  NOTICE: CAESinkAUDIOTRACK::Initializing with: m_sampleRate: 192000 format: AE_FMT_S16LE (AE) method: IEC (PT) stream-type: STREAM_TYPE_EAC3 min_buffer_size: 49216 m_frames: 6152 m_frameSize: 4 channels: 2
2019-10-05 00:08:55.475 T:12752   ERROR: CDVDVideoCodecAndroidMediaCodec::AddData error(-10000)
2019-10-05 00:10:04.474 T:12643   ERROR: Previous line repeats 5241 times.
2019-10-05 00:10:04.474 T:12643  NOTICE: NFS is idle. Closing the remaining connections.
2019-10-05 00:10:04.483 T:12752   ERROR: CDVDVideoCodecAndroidMediaCodec::AddData error(-10000)
2019-10-05 00:11:45.284 T:12643   ERROR: Previous line repeats 7640 times.
2019-10-05 00:11:45.284 T:12643  NOTICE: Whitelist search for: width: 1280, height: 720, fps: 23.000, 3D: false
[...]

PR #207, opened 2019-10-05
Add InputStream Helper information to logs

dagwieers:
Since making a screenshot is harder than copy&pasting logs, it is
probably better to log this information as well when it is requested.

This may help in troubleshooting.

This relates to #206


mediaminister:
Looks good to me.

PR #208, opened 2019-10-07
Add Widevine version as built-in

dagwieers:
Rather than leaving Widevine information out of the information pane, we
should display it as 'built-in'. This will be less confusing.

This relates to #206

### DIFF
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -1061,17 +1061,19 @@ class Helper:
         text += ' - ' + localize(30812, version=self._inputstream_version(), state=istream_state) + '\n'
         text += '\n'
 
-        if system_os() != 'Android':
-            text += localize(30820) + '\n'  # Widevine information
+        text += ' - ' + localize(30820) + '\n'  # Widevine information
+        if system_os() == 'Android':
+            text += ' - ' + localize(30821) + '\n'
+        else:
             from datetime import datetime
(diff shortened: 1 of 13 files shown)


horstle:
Instead of "Widevine Version: Built in" I think it would be better to add a sentence like "Widevine is built in" (maybe adding "on Android").
This could then also be added as a notification e.g. in `_has_widevine()` on line 336.

horstle:
I think we need to decide where and how the message should be shown. In my opinion:
- The API calls for install or removal should show it as a pop-up
- `_check_widevine()` and `_has_widevine()` should show it as a notification in the top right corner
- The info dialog should show it under "Widevine information"

The notification might need to be shorter than the other two. What exactly do you think should be the content?

dagwieers:
The most important aspect of inputstreamhelper is that it should stay out of the way. People shouldn't know it is there, unless there's a problem.

So to me, apart from what is already implemented, the only thing that is required from your list, is that we should give an ok-dialog with a message if people intend to install or remove Widevine explicitly (through API). In all other cases we should stay out of the way.

On Android it simply means: *we don't manage Widevine*, which means, no notifications, no means of installing/removing, nothing. Users shouldn't care.

horstle:
In the settings page the install and remove actions are already removed on android. I think it should be added to the docs that this should be done, too, wherever the API is used.
But we can't rely on them implementing this correctly, so I agree, that a message and return should be added there. The question now is should it be in __init__.py or already in api.py?

dagwieers:
I am going to merge this.
Adding a notification when using the API could be added at a later stage.

PR #216, opened 2019-10-30
Support undetected version

dagwieers:
For some reason unclear yet, I ended up with a widevinecdm.so that had
size 0, so obviosuly it didn't match the regexp and match would be None.

Resulting in:
```
File "/storage/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/__init__.py",
line 244, in _get_lib_version
  return match.group(0).lstrip('0')
AttributeError: 'NoneType' object has no attribute 'group'
```


mediaminister:
Looks good to me.

Issue #220, opened 2019-11-01
Commands fail with paths using spaces


beam:
ChromeOS recovery downloaded, extracted and mounted.
FDisk:
```
GPT PMBR size mismatch (8609948 != 4436015) will be corrected by w(rite).
Disk /media/Externi 500GB/Temp/temp/chromeos_12371.75.2_veyron-fievel_recovery_stable-channel_fievel-mp.bin: 2.1 GiB, 2271240192 bytes, 4436016 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: gpt
Disk identifier: A3C949A5-3203-4746-9CCF-C7D83632A57C

Device                                                                                                     Start     End Sectors  Size Type
/media/Externi 500GB/Temp/temp/chromeos_12371.75.2_veyron-fievel_recovery_stable-channel_fievel-mp.bin1  4415488 4435967   20480   10M Microsoft basic data
/media/Externi 500GB/Temp/temp/chromeos_12371.75.2_veyron-fievel_recovery_stable-channel_fievel-mp.bin2    20480   53247   32768   16M ChromeOS kernel
/media/Externi 500GB/Temp/temp/chromeos_12371.75.2_veyron-fievel_recovery_stable-channel_fievel-mp.bin3   319488 4415487 4096000    2G ChromeOS root fs
/media/Externi 500GB/Temp/temp/chromeos_12371.75.2_veyron-fievel_recovery_stable-channel_fievel-mp.bin4    53248   86015   32768   16M ChromeOS kernel
/media/Externi 500GB/Temp/temp/chromeos_12371.75.2_veyron-fievel_recovery_stable-channel_fievel-mp.bin5   315392  319487    4096    2M ChromeOS root fs
/media/Externi 500GB/Temp/temp/chromeos_12371.75.2_veyron-fievel_recovery_stable-channel_fievel-mp.bin6    16448   16448       1  512B ChromeOS kernel
/media/Externi 500GB/Temp/temp/chromeos_12371.75.2_veyron-fievel_recovery_stable-channel_fievel-mp.bin7    16449   16449       1  512B ChromeOS root fs
[...]

PR #224, opened 2019-11-09
support paths with spaces when calculating Chrome OS losetup offset

mediaminister:
This fixes https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/220

I tested this on a x86-64 Linux system and it should work for paths with or without spaces with `fdisk` and with `parted` both on `python2` and `python3`.

Not yet tested in a real life Kodi environment! Please test with different temporary download directories on different ARM systems with `parted` and `fdisk` before merging!


dagwieers:
We got a confirmation ;-)

Issue #210, opened 2019-10-13
Widevine CDM install crashing on ARM


GitUser4711:
During extraction (according to dialog) of the downloaded recovery image, process crashes. Log see attachement. Traced it down to the extraction run through _run_cmd(), but no idea where to go from there...

[error.txt](https://github.com/emilsvennesson/script.module.inputstreamhelper/files/3722098/error.txt)

InputStream Helper 0.4.3 (0.4.2 previously also failed)
Kodi 18.2 Leia from Raspbian Stretch official repository
Raspberry Pi 3b on Raspbian Stretch, latest updates/upgrades from official repository

Let me know in case you require additional information. Thanks!

GitUser4711:
I figured out the root cause. On a regular Linux system (so **not** LibreElec, OSMC, etc.), your user running Kodi might not have the right to use `sudo` command. In that case your agreement with the addon dialog ("plugin will use root commands x , y ,z") does not help and the related commands, mainly required for mounting the downloaded Chrome OS image, will fail.
You can work around this by adding the three commands mentioned in the respective addon dialog to your `etc/sudoers` file for the user running Kodi.

You might want to add a specific error message or at least remove the encoding issue from the current one so that the error message points in a more helpful direction.

mediaminister:
Unicode problems should be fixed in version 0.4.3, please provide a full debug log from version 0.4.3

mediaminister:
@GitUser4711 Can you try this fix: 
[script.module.inputstreamhelper-0.4.3-unicode-9b0d5c2.zip](https://github.com/emilsvennesson/script.module.inputstreamhelper/files/3848529/script.module.inputstreamhelper-0.4.3-unicode-9b0d5c2.zip)

dagwieers:
We did not get feedback. I assume this was fixed, if not reopen this issue, or leave feedback and we can reopen it.

PR #225, opened 2019-11-13
Use LANG=C for subprocesses

janhicken:
If `LANG=de_DE.UTF-8` is used, for example, the command output of `fdisk -l` will contain unicode characters. This is problematic because Kodi's logging mechanism crashes and the command's output parsing is not reliable.

When `LANG=C` is used, the commands' output will consist of ASCII-only characters, which is nice.

See #210 for an example of the resulting error message when a unicode string is given to the `log()` function. As every command output is given to `log()`, this means bad things.

### DIFF
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -198,7 +198,7 @@ class Helper:
         """Check whether cmd exists on system."""
         # https://stackoverflow.com/questions/377017/test-if-executable-exists-in-python
         import subprocess
-        return subprocess.call('type ' + cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE) == 0
+        return subprocess.call(['type ' + cmd], shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE) == 0
 
     def _update_temp_path(self, new_temp_path):
         """"Updates temp_path and merges files."""
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -279,13 +279,15 @@ class Helper:
         ''' Run subprocess command and return if it succeeds as a bool '''
         from .unicodehelper import to_unicode
         import subprocess
+        env = os.environ.copy()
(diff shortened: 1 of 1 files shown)


mediaminister:
Unicode problems should be fixed in version 0.4.3 and logging should work with all unicode characters.
This pull request works around a possible problem from a previous InputStreamHelper version, I'm not going to merge this.

If there is still a problem with version 0.4.3, please provide a full debug or make a new pull request which respects unicode support.

dagwieers:
I think it is worthwhile to force `LANG=C` in any case, it helps troubleshooting when command output is not translated and things become a bit more deterministic as we depend on reliable output.

That said, I hope we can rewrite this part of the code at some point.

mediaminister:
As this add-on supports multiple languages with different alphabets like Russian and Greek, I'm convinced supporting unicode everywhere is the best solution.
In specific cases I can allow exceptions, but I should first have a full debug log of this problem before I can even think about giving up unicode support.

janhicken:
Nobody's willing to drop unicode support anywhere, I think it's fine.

Still, the plain command output of `fdisk -l` for example is parsed in the code to retrieve some information out of it, for example the offset for the relevant partition to mount.

While my changes aren't necessary to fix the given problem in particular, it still increases the robustness of the parsing process itself. 

I did not have time to test if my problem still exists with the newest version of the plugin, but I'll report my findings in the next couple of days.

dagwieers:
@mediaminister This only affects the commands being run, so it should not matter for the rest of our add-on.

mediaminister:
I did a quick test on Linux with special chars and I still get `UnicodeDecodeError` with this.
Converting `subprocess` output to unicode will fix https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/210

dagwieers:
@mediaminister Using `LANG=C` for me is unrelated to Unicode-processing, the purpose is to ensure the output of commands are more predictable and (error) output is in English.

dagwieers:
I think we can merge this one as well.

Thanks @janhicken !

PR #233, opened 2019-11-27
Remove id from settings entries

dagwieers:
We currently have the id attribute added to entries in the settings.xml
that don't need it. And this results in entries in the user's settings
for things that don't store values.

So we ought to remove them to keep the user settings clean (for new users).


dagwieers:
I am not going to wait with merging this, as this is basically a no-brainer.

PR #235, opened 2019-12-08
Updated german translation

tweimer:


resources/language/resource.language.de_de/strings.po
```diff
@@ -307,4 +315,12 @@ msgstr "Widevine CDM Bibliothek (erneut) installieren..."
...
-msgstr "Entferne die Widevine CDM Bibliothek..."
+msgstr "Widevine CDM Bibliothek entfernen..."
+
+msgctxt "#30913"
+msgid "Number of backups"
+msgstr "Anzahl Backups"
+
+msgctxt "#30915"
+msgid "Rollback Widevine CDM library..."
+msgstr "Widevine CDM Bibliothek zurücksetzen..."
```
horstle: I think "zurückrollen" would be a better translation of "rollback" than "zurücksetzen", but actually I'd like to discuss, whether "rollback" is actually the correct word here.
It fits at the moment, but in the future, when the installation of versions not backed up locally is added, it might be that this option actually installs a newer version, while rollback implies installing an older version. So should that be changed to sth. like "Change version of Widevine CDM library..."?

tweimer: Well, doesn't it actually revert to the previous version, if the current one cannot be installed for any reason? I never heard or used "zurückrollen" in any technical context.

horstle: You don't put the library back to its previous state, but replace it with a previous (or any different) version. IMO "zurücksetzen" would mean there were modifications done to the library itself which are now reverted. I also thought "zurückrollen" is not ideal, but since you also say "ausrollen", when publishing a new version, I thought it would be understandable.

Maybe "ersetzen" would fit better, but actually I'd like to postpone this discussion until it's decided whether "rollback" will be changed or not.
By the way I completely agree with all the rest of your changes/additions.

What do you think about "rollback" @emilsvennesson  @dagwieers @mediaminister? Is it the correct word here? Even now you can already change to an older version in this context and then back to a newer one.

tweimer: From my understanding, **zurückrollen** and **zurücksetzen** mean exactly the same thing, while the last one is more commenly used IMHO.

horstle: No, "zurücksetzen" means to put something back to its original state. That's just not what is done here.

tweimer: Well, it is the same meaning as "zurückrollen" ( if I assume the englisch phrase is correct).

dagwieers: You guys basically have to agree, or we need a tiebreaker ;-)

mediaminister: Maybe changing it in English to "restore" and then use "wiederherstellen" in German?

tweimer: @mediaminister Yes, I agree with that.

horstle: Fine by me.

tweimer:
@dagwieers @horstle I have changed that for the german translation - do you change the rest of them?

tweimer:
Updated translation for latest changes.

PR #240, opened 2019-12-22
Support vendor widevine library

### LINKED ISSUE #229: Support vendor distributed Widevine
Hi,

I reached out to @dagwieers initially via email as this isn't an issue per se. 

It would be good to start a discussion about how to handle vendor distributed versions of Widevine. We (OSMC) would like to distribute our own version of Widevine. We have satisfied the legal requirements to distribute these libraries directly for some supported platforms. 

Understandably, this add-on is brought in as a dependency by a number of add-ons that utilise Widevine. I'd suggested a symlink (such as /usr/lib/libwidevine-cdm.so -> /usr/lib/libwidevine-vendor.so to signal recognition), but @dagwieers has mentioned quite correctly that this wouldn't be suitable for Windows.

In short, it would be good to adapt this add-on so it can detect a vendor provided version of Widevine and devolve updates to the distribution. 

Cheers

Sam

dagwieers:
This adds support for a vendor-supported Widevine library.

**TODO:**
- Show actual version, not stored version
- Update stored version
- Get timestamp from library to show in info pane
- Do not show Chrome OS information when using vendor library
- Disable settings/operations

This fixes #229

### DIFF
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -1074,7 +1074,7 @@ class Helper:
         if system_os() == 'Android':
             text += ' - ' + localize(30821) + '\n'
         elif self._has_vendor_widevine():
-            text += ' - ' + localize(30822) + '\n'
+            text += ' - ' + localize(30822, version=self._get_lib_version(self._widevine_path())) + '\n'
         else:
             from datetime import datetime
             wv_updated = datetime.fromtimestamp(float(get_setting('last_update'))).strftime("%Y-%m-%d %H:%M") if get_setting('last_update') else 'Never'
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -3,8 +3,10 @@
 from __future__ import absolute_import, division, unicode_literals
 import os
 from inputstreamhelper import config
-from .kodiutils import (addon_profile, browsesingle, get_addon_info, get_proxies, get_setting, jsonrpc, kodi_to_ascii, localize,
-                        log, notification, ok_dialog, progress_dialog, select_dialog, set_setting, textviewer, translate_path, yesno_dialog)
+from .kodiutils import (addon_profile, browsesingle, copy, delete, exists, get_addon_info, get_proxies, get_setting,
+                        hardlink, jsonrpc, kodi_to_ascii, localize, log, mkdir, mkdirs, notification, ok_dialog,
+                        progress_dialog, samefile, select_dialog, set_setting, textviewer, translate_path,
+                        yesno_dialog)
 
 # NOTE: Work around issue caused by platform still using os.popen()
 #       This helps to survive 'IOError: [Errno 10] No child processes'
@@ -46,7 +48,7 @@ class Helper:
         self.drm = drm
 
         from platform import uname
-        log('Platform information: {uname}', uname=uname())
+        log(0, 'Platform information: {uname}', uname=uname())
 
         if self.protocol not in config.INPUTSTREAM_PROTOCOLS:
             raise InputStreamException('UnsupportedProtocol')
@@ -81,7 +83,6 @@ class Helper:
     @classmethod
     def _temp_path(cls):
         """Return temporary path, usually ~/.kodi/userdata/addon_data/script.module.inputstreamhelper/temp"""
-        from xbmcvfs import exists, mkdirs
         temp_path = translate_path(os.path.join(get_setting('temp_path', 'special://masterprofile/addon_data/script.module.inputstreamhelper'), 'temp'))
         if not exists(temp_path):
             mkdirs(temp_path)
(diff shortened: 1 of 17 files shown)


dagwieers:
The information pane will look like this:
> **Kodi information**
> - Kodi version: **18.5**
> - Kodi platform/architecture: **Linux/arm**
>
> **InputStream information**
> - InputStream Helper version: **0.4.3**
> - InputStream Adaptive version: **2.4.0**
>
> **Widevine information**
>- Widevine version: **4.10.1440.18 (supplied by vendor)**
>
> Please report issues to:
> - https://github.com/emilsvennesson/script.module.inputstreamhelper/issues

dagwieers:
We need to test this functionality on Windows and macOS, mostly to test the behavior using hardlinking and exception handling.

dagwieers:
@horstle Hardlinking here is not just useful to save disk space, but it makes testing if two files are the same file a lot more efficient than the worst case scenario: comparing complete content. Especially if this has to happen on every playback...

horstle:
I don't have any problem with (hard)links. I just prefer avoiding/minimising differences in behaviour on different platforms, but let's not start that discussion again.

dagwieers:
@horstle Oh, I didn't want to start a polemic, I just found a better reason to use them (when they are supported)

dagwieers:
@samnazarko Would it be useful to show the library timestamp as well in the information pane (i.e. ctime or mtime) so the user has an indication when the library was made/added?

lib/inputstreamhelper/__init__.py
```diff
@@ -146,7 +158,15 @@ def _widevine_path(cls):
...
-            from xbmcvfs import exists
+
+            # Support vendor-supplied Widevine CDM
+            vendor_widevine_path = '{0}_vendor{1}'.format(*os.path.splitext(widevine_path))
+            if exists(vendor_widevine_path):
+                if exists(widevine_path):
+                    if not samefile(vendor_widevine_path, widevine_path):
+                        hardlink(vendor_widevine_path, widevine_path)
+                else:
+                    hardlink(vendor_widevine_path, widevine_path)
```
horstle: I think it made more sense if `_widevine_path()` only did what its name suggests, i.e. return the path to the Widevine lib.
If these lines were moved to `_has_vendor_widevine()` and a check for `_has_vendor_widevine()` was added to `_has_widevine()` I think it would still work and in my opinion would be more logical.

dagwieers: @horstle I don't disagree, but there's more duplicate code if I split it off :disappointed: I implemented now what you asked for.

lib/inputstreamhelper/__init__.py
```diff
@@ -137,21 +135,59 @@ def _load_widevine_config(cls):
...
+    def _hardlink_vendor_widevine(cls):
+        """Hardlink vendor Widevine into place"""
+        widevine_cdm_filename = config.WIDEVINE_CDM_FILENAME[system_os()]
+        if widevine_cdm_filename is None:
+            return
+
+        if cls._ia_cdm_path() is None:
+            return
+
+        widevine_path = os.path.join(cls._ia_cdm_path(), widevine_cdm_filename)
```
horstle: You could shorten lines 158-165 by using `_widevine_path()`.

dagwieers: Yeah, I didn't like the interface, but I also did not want to rewrite too many things either as before you know it I am rewriting half of the add-on :-( Been there, done that.

dagwieers: Alright, so here is that rewrite ;-) I had to split up `_widevine_path()` into two functions `_has_widevinecdm()` and `_widevinecdm_path()`. And did the same for `_has_vendor widevinecdm()` and `_vendor_widevinecdm_path()`.

dagwieers: It is not exactly shorter though.

samnazarko:
I don’t think it’s necessary to show any significant versioning details for vendor based Widevine libs.

I’ll send our current version to your email shortly so you can have a look


On 23 Dec 2019, at 11:03, Dag Wieers <notifications@github.com> wrote:

﻿

@samnazarko<https://github.com/samnazarko> Would it be useful to show the library timestamp as well in the information pane (i.e. ctime or mtime) so the user has an indication when the library was made/added?

dagwieers:
@samnazarko Without versioning information, how would users relate issues to versions/updates. Especially in the light of regressions (as have been reported in the past).

samnazarko:
From a distribution perspective, we would expect a user to say it worked in the January version of OSMC but broke / regressed in February. Our release notes would show a version bump.

Our lib is stable and is somewhat pegged to CDM updates. I know there have been regressions with SW decoding but don’t anticipate this for HW decoding - it either works or it doesn’t.

On 24 Dec 2019, at 19:41, Dag Wieers <notifications@github.com> wrote:

﻿

@samnazarko<https://github.com/samnazarko> Without versioning information, how would users relate issues to versions/updates. Especially in the light of regressions (as have been reported in the past).

samnazarko:
I've emailed through our current Widevine package.

dagwieers:
I had to rework this PR for a big code restructure and prefered to start from scratch while keeping the original branch. So there is now a new PR #311 with the necessary changes for vendor widevine support.

Most of the non-vendor related changes in this PR were recently merged as part of #310 

And we can close this one.

Issue #265, opened 2020-02-08
Error while extracting chrome os recovery image


Chris2711:
Hello everybody,
while setting up Netflix-Addon ( CastagnaIT / plugin.video.netflix)  to my Raspi3+ (libreelec 9.2.0) everything is working fine until downloading the chrome OS recovery images. When download is complete i get following error while extracting the package :

error: Error -3 while decompressing: invalid stored block lengths

I´ve installed:
plugin.video.netflix-0.16.4.zip / CastagnaIT

and tried following modules due to #146 (unable update or re-install widevine for netflix):
script.module.inputstreamhelper-0.4.1-armfix-5dc3e7b.zip
script.module.inputstreamhelper-0.4.2-release-73fc720.zip
and the actual one 0.4.3 which is automatically installed with the plugin

What´s wrong here. I attached my kodi.log. Can anybody help me?

Thanks Chris
[KODI LOG.txt](https://github.com/emilsvennesson/script.module.inputstreamhelper/files/4177850/KODI.LOG.txt)

dcambria:
Similar problem here. Tested with script.module.inputstreamhelper-0.4.2 and 0.4.3 from repo.
Thanks
[kodi.log](https://github.com/emilsvennesson/script.module.inputstreamhelper/files/4184186/kodi.log)


2020-02-11 02:04:19.448 T:1475338464   DEBUG: [script.module.inputstreamhelper-0.4.2]: [u'modinfo', u'loop'] cmd failed.
2020-02-11 02:04:19.455 T:1475338464   DEBUG: [script.module.inputstreamhelper-0.4.2]: [u'modinfo', u'loop'] cmd output:
                                            modinfo: ERROR: Module loop not found.
2020-02-11 02:04:19.456 T:1475338464   DEBUG: [script.module.inputstreamhelper-0.4.2]: loop is built in the kernel.
2020-02-11 02:04:19.556 T:1475338464   DEBUG: [script.module.inputstreamhelper-0.4.2]: [u'losetup', u'-f'] cmd executed successfully.
2020-02-11 02:04:19.557 T:1475338464   DEBUG: [script.module.inputstreamhelper-0.4.2]: [u'losetup', u'-f'] cmd output:
                                            /dev/loop1
2020-02-11 02:04:19.557 T:1475338464   DEBUG: [script.module.inputstreamhelper-0.4.2]: Found free loop device: /dev/loop1
[...]

PR #277, opened 2020-02-26
Add option to calculate checksum in _http_download()

horstle:
I tested the following script on an RPi3 using Raspbian:

```python
from hashlib import md5, sha1
from datetime import datetime

def check(alg):
    with open("chromeos_12607.82.0_nyan-kitty_recovery_stable-channel_kitty-mp-v2.bin.zip", "r") as f:
        starttime = datetime.now()
        print("{} started at {}".format(alg, starttime))
        if alg == 'md5':
            checksum = md5()
            should_be = '245b4eb00dba6752b0038ba447440799'
        elif alg == 'sha1':
            checksum = sha1()
            should_be = '86afee8b24784947cfecbc49a42fe58538a003b5'
        else:
            print('Specify sha1 or md5')
            return

#        chunktime = datetime.now()
        while True:
            chunk = f.read(32*1024)
            if not chunk:
                break
            checksum.update(chunk)
#            print("next chunk calculated after {}, total {}".format(datetime.now() - chunktime, datetime.now() - starttime))
#            chunktime = datetime.now()
        h = checksum.hexdigest()
        print("Calculated {} with hexdigest {} at {} after {}".format(checksum, h, datetime.now(), datetime.now() - starttime))
        if h == should_be:
            print('Checksum correct')
        else:
            print('CHECKSUM INCORRECT!')

check('md5')
check('sha1')
check('md5')
check('sha1')
```

The results were 3.7s for md5 and 4.7s for sha1 (both times). I think that's fast enough to be negligible compared to download time.
So this PR adds the option to calculate the checksum to `_http_download()` and use it in `_install_widevine_arm()`.

This relates to #265.

### DIFF
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -398,7 +398,8 @@ class Helper:
             elif hash_alg == 'md5':
                 calc_checksum = md5()
             else:
-                raise ValueError('Invalid hash algorithm specified: {}'.format(hash_alg))
+                log('Invalid hash algorithm specified: {}'.format(hash_alg))
+                checksum = None
 
         req = self._http_request(url)
         if req is None:
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -389,8 +389,17 @@ class Helper:
         # log('Response: {response}', response=content)
         return content.decode()
 
-    def _http_download(self, url, message=None):
+    def _http_download(self, url, message=None, checksum=None, hash_alg='sha1'):
         """Makes HTTP request and displays a progress dialog on download."""
+        if checksum:
+            from hashlib import sha1, md5
+            if hash_alg == 'sha1':
+                calc_checksum = sha1()
+            elif hash_alg == 'md5':
+                calc_checksum = md5()
+            else:
+                raise ValueError('Invalid hash algorithm specified: {}'.format(hash_alg))
+
         req = self._http_request(url)
         if req is None:
             return None
@@ -412,6 +421,8 @@ class Helper:
                 if not chunk:
                     break
                 image.write(chunk)
+                if checksum:
+                    calc_checksum.update(chunk)
                 size += len(chunk)
                 percent = int(size * 100 / total_length)
                 if progress.iscanceled():
@@ -420,6 +431,9 @@ class Helper:
                     return False
                 progress.update(percent)
 
+        if checksum and not calc_checksum.hexdigest() == checksum:
+            log('Download failed, checksums do not match!')
+            return False
         progress.close()
         req.close()
         return True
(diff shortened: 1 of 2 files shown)


dagwieers:
@horstle Very good that you do this during download, which avoids having to read it again from disk. So well done. However, we don't know if all chunks were written to disk. And for that reason I think we still ought to check if the final file size is exactly as we expect.

lib/inputstreamhelper/__init__.py
```diff
@@ -389,8 +389,17 @@ def _http_get(self, url):
...
+    def _http_download(self, url, message=None, checksum=None, hash_alg='sha1'):
         """Makes HTTP request and displays a progress dialog on download."""
+        if checksum:
+            from hashlib import sha1, md5
+            if hash_alg == 'sha1':
+                calc_checksum = sha1()
+            elif hash_alg == 'md5':
+                calc_checksum = md5()
+            else:
+                raise ValueError('Invalid hash algorithm specified: {}'.format(hash_alg))
```
dagwieers: Not sure if we want this to blow up like this. Since it now works without checksum verification, maybe we ought to just report it, but still continue without verification.

horstle: Ok, changed it.

dagwieers:
# [Codecov](https://codecov.io/gh/emilsvennesson/script.module.inputstreamhelper/pull/277?src=pr&el=h1) Report
> Merging [#277](https://codecov.io/gh/emilsvennesson/script.module.inputstreamhelper/pull/277?src=pr&el=desc) into [master](https://codecov.io/gh/emilsvennesson/script.module.inputstreamhelper/commit/262dd884e8a247229f9e935de3c12361f65bcefd&el=desc) will **decrease** coverage by `0.17%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/emilsvennesson/script.module.inputstreamhelper/pull/277?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master     #277      +/-   ##
==========================================
- Coverage   78.06%   77.89%   -0.18%     
==========================================
  Files          18       18              
  Lines        1819     1832      +13     
==========================================
+ Hits         1420     1427       +7     
- Misses        399      405       +6     
```



------

[Continue to review full report at Codecov](https://codecov.io/gh/emilsvennesson/script.module.inputstreamhelper/pull/277?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/emilsvennesson/script.module.inputstreamhelper/pull/277?src=pr&el=footer). Last update [262dd88...ab2bd23](https://codecov.io/gh/emilsvennesson/script.module.inputstreamhelper/pull/277?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

horstle:
I can probably add another commit (or maybe another PR?) for file size check sometime on friday.

horstle:
Alright, checking file size is added, I tested it on an RPi4, where the download actually took 2s longer with current master than with this change (must be due to inconsistent download speed).

dagwieers:
Even if there would be a 2 second overhead due to this verification, that would be totally worth it IMO.

dagwieers:
And I need to fix codecov :disappointed:

horstle:
Thanks!

PR #287, opened 2020-03-04
Use explicit relative imports instead of sys.path.append

mediaminister:
Many thanks to @enen92 for figuring this solution out.
Meanwhile I tested this on the following systems and it just works fine!
- Kodi 19.0-ALPHA1 Git:20200226-cde760b94e on Windows x64
- Kodi 18.6 on Linux x64 and Windows x64
- LibreELEC 9.2.0
- LibreELEC 8.2.5

### DIFF
--- a/default.py
+++ b/default.py
@@ -3,18 +3,6 @@
 
 from __future__ import absolute_import, division, unicode_literals
 import sys
-import os.path
-from xbmc import translatePath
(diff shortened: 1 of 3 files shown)


horstle:
LGTM. Unit tests seem to run fine. I'd merge this right now if I could.

PR #290, opened 2020-03-06
Change when last_update is set

### LINKED ISSUE #191: 'last_update' not always updated on ARM
I didn't test this yet, but I think if somebody makes a reinstall via the settings page on ARM, it wouldn't update the 'last_update' setting, since this is only done by `_latest_widevine_version()` which isn't called then (only in `_install_widevine_x86()` or `_update_widevine()`).

So I think the line 
```
set_setting('last_update', str(time.mktime(datetime.utcnow().timetuple())))
```
should be moved to `install_widevine()` .

That way on ARM, as on x86, after you cancel an install it would not ask again immediately. At first I thought this relates to @the-isz 's problem (see [here](https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/165#issuecomment-533892308)) and makes it extremely annoying, but in his case `_update_widevine()` is probably executed?

BTW the default for 'update_frequency' in `_update_widevine()` is still set to 14 days. Even though this is only relevant for tests, I think it should be changed to the real default value to be consistent.

horstle:
This fixes #191 .

I did a quick test on Linux x86. Seems to work fine. Since it's not a big change I don't really expect any issues on ARM either.

### DIFF
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -542,9 +542,6 @@ class Helper:
             versions = self._http_get(url)
             return versions.split()[-1]
 
-        from datetime import datetime
-        from time import mktime
-        set_setting('last_update', mktime(datetime.utcnow().timetuple()))
         if 'x86' in self._arch():
             url = config.WIDEVINE_VERSIONS_URL
             versions = self._http_get(url)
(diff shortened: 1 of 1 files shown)


mediaminister:
LGTM.

PR #291, opened 2020-03-06
Restructure the code

horstle:
I moved `http_request`, `http_get`, `http_download`, `temp_path` and `update_temp_path` to utils.py, with some necessary changes. Hopefully I managed not to break any other PR.
At first I wanted to keep `from_unicode` and `to_unicode` in utils, but that would have resulted in circular imports between utils and kodiutils.

Maybe as a next step `unzip` (and `widevine_eula`?) could be moved to utils, too. Then there might be a better solution for `self._download_path`.

This relates to #94 .


mediaminister:
@horstle I moved `unzip` to `utils`, I think we should leave all widevine related code in one file, so I didn't move `widevine_eula`.

Please test, so this can be merged.

horstle:
Actually I meant to move more functions in another PR, so each PR won't be too big, but that's ok. I did some tests and it seems to work well. Could be merged IMO.

PR #293, opened 2020-03-11
Support exporting Kodi 19 Matrix compatible zip package

mediaminister:
This adds support for automatically building a Kodi 19 Matrix compatible zip package with `make`.

This includes:
- automatically change the Python ABI version in addon.xml (change to **3.0.0** for Matrix)
- automatically change the addon version using local version identifier ([PEP-440](https://www.python.org/dev/peps/pep-0440/#local-version-identifiers)) (add **+matrix.1** to current version for Matrix)

How to build:
```
cd script.module.inputstreamhelper
make multizip
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,12 @@
 export PYTHONPATH := $(CURDIR)/lib:$(CURDIR)/test
 PYTHON := python
+KODI_PYTHON_ABIS := 3.0.0 2.25.0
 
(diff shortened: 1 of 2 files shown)


dagwieers:
That's looking good, I haven't tested it.
It does look quite complex though, I had something more simple in mind, but this reuses `make zip` better :-)

PR #297, opened 2020-03-25
Add spanish translation [es_ES]

roliverosc:


mediaminister:
Thanks for the translation!

PR #298, opened 2020-03-27
Add compatibility for Kodi 19 Python API

mediaminister:
This pull request adds compatibility with Kodi 19 Python API while preserving backwards compatibility with earlier Kodi versions.

### DIFF
--- a/lib/inputstreamhelper/kodiutils.py
+++ b/lib/inputstreamhelper/kodiutils.py
@@ -5,10 +5,27 @@
 from __future__ import absolute_import, division, unicode_literals
 import xbmc
 import xbmcaddon
+from xbmcgui import DialogProgress
 from .unicodes import from_unicode, to_unicode
 
 # NOTE: We need to add the add-on id in here explicitly !
 ADDON = xbmcaddon.Addon('script.module.inputstreamhelper')
+
+
+class progress_dialog(DialogProgress, object):  # pylint: disable=invalid-name,useless-object-inheritance
+    """Show Kodi's Progress dialog"""
+
+    def create(self, heading, message=''):  # pylint: disable=arguments-differ
+        """Create and show a progress dialog"""
+        if kodi_version_major() < 19:
+            return super(progress_dialog, self).create(heading, line1=message)
+        return super(progress_dialog, self).create(heading, message=message)
+
+    def update(self, percent, message=''):  # pylint: disable=arguments-differ
+        """Update the progress dialog"""
+        if kodi_version_major() < 19:
+            return super(progress_dialog, self).update(percent, line1=message)
+        return super(progress_dialog, self).update(percent, message=message)
 
 
 class SafeDict(dict):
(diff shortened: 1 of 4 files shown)


test/xbmcgui.py
```diff
@@ -110,38 +112,41 @@ class DialogProgress:
...
-    def create(self, heading, line1, line2=None, line3=None):
+    def create(self, heading, message='', line1='', line2='', line3=''):
         """A stub implementation for the xbmcgui DialogProgress class create() method"""
-        self.percentage = 0
+        self.percent = 0
         heading = kodi_to_ansi(heading)
+        message = kodi_to_ansi(message)
         line1 = kodi_to_ansi(line1)
-        print('\033[37;44;1mPROGRESS:\033[35;49;1m [%s] \033[37;1m%s\033[39;0m' % (heading, line1))
+        message = kodi_to_ansi(message)
```
horstle: Why did you add this twice?

lib/inputstreamhelper/kodiutils.py
```diff
@@ -106,7 +119,9 @@ def yesno_dialog(heading='', message='', nolabel=None, yeslabel=None, autoclose=
     from xbmcgui import Dialog
     if not heading:
         heading = ADDON.getAddonInfo('name')
-    return Dialog().yesno(heading=heading, line1=message, nolabel=nolabel, yeslabel=yeslabel, autoclose=autoclose)
+    if kodi_version_major() < 19:
+        return Dialog().yesno(heading=heading, line1=message, nolabel=nolabel, yeslabel=yeslabel, autoclose=autoclose)
```
dagwieers: We should probably move this to using kodiutils as well.

test/xbmcgui.py
mediaminister: That is a typo. Thanks for noticing.

lib/inputstreamhelper/kodiutils.py
mediaminister: I don't understand, what do you mean exactly?

dagwieers: The kodiutils library is an abstraction layer to all Kodi-related calls, and here we aren't using it.

mediaminister: I'm confused, this code is in **kodiutils.py**

horstle: @dagwieers You probably didn't see this is inside kodiutils?

dagwieers: Ok, I was confused. I thought we moved everything to kodiutils in the past, so I was surprised to see this :smile: Move along, nothing to see here...

dagwieers:
@mediaminister Since this has been confirmed to be working, and this is conflicting with my GitHub CI work, I am going to merge this now. If there are additional fixes needed later, we can still add them later. Sorry for the inconvenience.

dagwieers:
I was wondering what the overhead was of using `inspect.getargspec()` instead of a version check.

It would support both older Kodi v19 releases as well as newer Kodi v19 releases (during the transition period). Some users do run e.g. LibreELEC test releases that have older Kodi versions, and are not easily updated. (See #338)

Issue #294, opened 2020-03-15
Coercing to unicode Error


spikerguy:
Bug report
Your Environment
Netflix add-on version: 1.0
Operating system version/name: Manjaro Linux - Arch - Aarch64
Device model: Pine64 PineBook Pro
Used Operating system:

 Linux - Aaarch64
Describe the bug
Coercing to unicode error while uncompressing Widevine CDM

Expected behavior
After uncompressing it tries to run some python package and fails, Assuming I have python2.7 and python3.8 installed.

Your Environment
Manjaro Linux with Kde Plasma compiled kodi for aarch64 using gles and panfrost driver. Video playback is quite smooth but trying to get netflix to work. 64bit userspace is used.

Steps to reproduce the behavior
Installed plugin.video.addon
Logged in to Netflix and tried playing a movie, this ask for permissions to download Chrome OS image and then after downloading it tries to uncompress it.
While uncompressing I get the Coercing to unicode error and the uncompressing box stays on until I close kodi.
Possible fix
Looks like a python issue.

Debug log
The debug log can be found from this link:
https://pastebin.com/6zb7S3G1

Installation
Installed this plugin from the git wiki.

Screenshots
**Uncompressing**
![image](https://user-images.githubusercontent.com/48839029/76698791-0741f480-66b8-11ea-8e97-dfcaa86bfd51.png)

**IMG_Coering**
![image](https://user-images.githubusercontent.com/48839029/76698790-03ae6d80-66b8-11ea-9d82-90a81b81767a.png)

mediaminister:
It seems you provided just a simple standard log, without specific debug log messages from InputSteamHelper. I can't investigate where it goes wrong. Can you please provide a **full debug log**, following the instructions here: https://kodi.wiki/view/Log_file/Easy

horstle:
This could possibly be related to #253 since it fails at the same point, although with a different error.

spikerguy:
I will try to get the detailed log by tomorrow.  Have anyone tested Netflix on 64bit userspace? 

Afaik it needs 32bit userspace for the plugin to make use of chromeos binaries.

spikerguy:
Hello,
Here is the logs.
https://paste.kodi.tv/poliyecaqo.kodi

horstle:
Ok, so `_set_loop_dev()` fails with permission denied, so `_loop_dev` is None, resulting in `_losetup()` crashing.

[...]

PR #302, opened 2020-03-31
Prepare restructure

horstle:
These commits prepare the code for restructuring.
* Class attributes are changed to function attributes. That function can be imported where ever needed, so it works across files.
* `_widevine_eula` and the `_install_widevine*`s are changed to make sure the arch-specific functions don't need generic widevine functions. (That would create circular imports when restructuring.)

This also fixes what I mentioned in https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/294#issuecomment-600465581


mediaminister:
Looks good to me.

PR #303, opened 2020-04-03
Add Romanian translation

tmihai20:
Translated strings into Romanian for the first time.


mediaminister:
Thanks for the translation!

PR #304, opened 2020-04-06
Restructure

horstle:
I think this change can be merged rather quickly, since it really only moves functions somewhere else, imports them and renames stuff where necessary.


mediaminister:
I tested and it works like before on Kodi 18 and Kodi 19. So, I'm going to merge it.

PR #308, opened 2020-04-11
Simplify userspace detection on ARM64

mediaminister:
This fixes https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/294

It is confirmed this works to detect  64-bit userspace on ARM64.

Before merging, this should be tested more, especially on Android.


dagwieers:
I broke this PR :-( But to me this is ready to be merged.

horstle:
I just tested this on an arm64 Android phone and it works as expected.

PR #309, opened 2020-04-11
Assorted fixes

dagwieers:
This PR includes:
- Remove items from .gitattributes
- Remove WHITEIP from ARM HWID list (it's Intel Apollo Lake)
- Fix wrong URL in setup.py


.gitattributes
```diff
@@ -1,10 +1,8 @@
 .github/ export-ignore
-dist/ export-ignore
 tests/ export-ignore
 .gitattributes export-ignore
 .gitignore export-ignore
 .pylintrc export-ignore
-codecov.yml export-ignore
```
mediaminister, line 7: I don't understand, does this mean `codecov.yml` should be exported now in zip packages?

michaelarnauts: It think the file is removed from the repo, and the settings are defined on the organisation.

dagwieers: Yes, I forgot to clean it up earlier.

Having codecov defined  at the org lrvel has one benefit, it doesn't suddenly fail to work. But if there is a need to have to tweak it, we can still reintroduce it.

mediaminister: There is still a codecov.yml in this repo!

dagwieers:
:roll_eyes: Fixed.

PR #312, opened 2020-04-14
Don't fail on untranslated strings

dagwieers:
Currently CI would fail on untranslated strings, but during development
this is to be expected. So we check now for both, but don't fail for
untranslated strings.


dagwieers:
Kodi add-on checker is currently broken for Krypton and Leia :-(
And we can't use an earlier release either.

PR #313, opened 2020-04-14
Cache architecture information

dagwieers:
Since we call `arch()` all over the place, it is better to get it once,
report on it and cache the value for future use.


horstle:
lgtm

mediaminister:
@dagwieers I think this breaks ARM tests.

mediaminister:
I made a pull request that should fix this caching problem with unit tests: https://github.com/emilsvennesson/script.module.inputstreamhelper/pull/322

PR #314, opened 2020-04-14
Make functions cache values

dagwieers:
I am not fond of the new `store()` functionality as it requires a
synchronous approach to where specific cached information is set and
used.

And I understand how it can help with the image-processing
functionality, however I think this mostly is a display of bad practices
and actually requires a redesign of the progress-bar and how we call each phase.

This PR returns back to the original "if cached, return cache, otherwise
perform lookup" for a simple case, but I'd like to see a different
approach to the other uses.

### DIFF
--- a/lib/inputstreamhelper/utils.py
+++ b/lib/inputstreamhelper/utils.py
@@ -147,17 +147,18 @@ def unzip(source, destination, file_to_unzip=None, result=[]):  # pylint: disabl
 def system_os():
     """Get system platform, and remember this information"""
 
-    # If it wasn't stored before, get the correct value
-    if not store('system_os'):
-        from xbmc import getCondVisibility
-        if getCondVisibility('system.platform.android'):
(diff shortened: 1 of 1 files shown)


horstle:
The store() function is not worse than the class attributes it replaced. Although I agree that it should be replaced in the long run, I don't think this is a priority.
Funny enough, the place where your reversion avoids the store function is the only situation, where the issue you mentioned does not apply. Still, I think this change can be merged, since there is no benefit in using store().
I'm currently working on the udisks and p7zip options for ARM, simultaneously removing calls to store() at some points.

dagwieers:
No it is not a priority, but I was doing it for arch() and decided to make a separate PR for this. Only then I saw the other cases were an exception and not exactly lookup functions. That is how I ended up with only this 😉

dagwieers:
So also not very useful as a draft.

PR #315, opened 2020-04-14
Add a 20 MiB buffer to required disk space

mediaminister:
Adding 20 MiB buffer to the required disk space won't hurt and possibly fixes https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/281

### DIFF
--- a/lib/inputstreamhelper/widevine/arm.py
+++ b/lib/inputstreamhelper/widevine/arm.py
@@ -162,7 +162,8 @@ def install_widevine_arm(backup_path):  # pylint: disable=too-many-statements
         log(4, 'We could not find an ARM device in the Chrome OS recovery.conf')
         ok_dialog(localize(30004), localize(30005))
         return ''
-    required_diskspace = int(arm_device['filesize']) + int(arm_device['zipfilesize'])
+    # Estimated required disk space: takes into account an extra 20 MiB buffer
+    required_diskspace = 20971520 + int(arm_device['zipfilesize']) + int(arm_device['filesize'])
     if yesno_dialog(localize(30001),  # Due to distributing issues, this takes a long time
                     localize(30006, diskspace=sizeof_fmt(required_diskspace))):
         if system_os() != 'Linux':


dagwieers:
10% of 2GB is quite a lot. I would add 10MB or 20MB for additional writes/logging.

mediaminister:
I changed it to 5%.

We also didn't take into account the EULA extraction from the x86 zip -> + 5 MB

And there is definitely a significant filesystem overhead on some systems:
https://rwmj.wordpress.com/2009/11/08/filesystem-metadata-overhead/
https://unix.stackexchange.com/questions/278647/overhead-of-using-loop-mounted-images-under-linux

So, at least 5%

lib/inputstreamhelper/widevine/arm.py
```diff
@@ -162,7 +162,8 @@ def install_widevine_arm(backup_path):  # pylint: disable=too-many-statements
         log(4, 'We could not find an ARM device in the Chrome OS recovery.conf')
         ok_dialog(localize(30004), localize(30005))
         return ''
-    required_diskspace = int(arm_device['filesize']) + int(arm_device['zipfilesize'])
+    # Estimated required disk space: take into account up to 5% filesystem overhead
+    required_diskspace = round(1.05 * int(arm_device['zipfilesize']) + int(arm_device['filesize']))
```
horstle, line 166: This would only add 5% of zipfilesize, not 5% of the sum. Is that on purpose?

mediaminister: No, I'll add parentheses

dagwieers:
> We also didn't take into account the EULA extraction from the x86 zip -> + 5 MB

Good catch!

> And there is definitely a significant filesystem overhead on some systems:

Sure there is overhead, in caching and CPU, but not additional file system use. Sure there may be additional paging going to disk, but that is already allocated space on disk (swap), not file system space.

Even 5% is close to 100MB additional disk space, which on smaller devices may be a dealbreaker. (If 100MB is the difference between a go and a no go, I agree that's not a good situation to be in)

I don't think it should be relative to compressed image + image size.

mediaminister:
I can't test on RPi, but it would help if someone could measure the overhead.

dagwieers:
Take 20MB, or if you want to be safe maybe 30MB. It's a timespan of about 3 to 4 minutes of logging, plus the EULA.

mediaminister:
Okay, 30 MiB it is.

dagwieers:
Just FYI, the link you shared (https://rwmj.wordpress.com/2009/11/08/filesystem-metadata-overhead/) about metadata overhead is an interesting one, but does not apply to a read-only image. If you mount an empty image and would write to a filesystem there is indeed file system metadata and a journal to take into account (depending on the filesystem used).

But in our case the image is fixed and the size is already allocated. What's more the uncompressed image may use more disk space than is actually used because even unused file system space will be allocated even when unused. (An empty 1TB image compresses well, but still is 1TB in size on disk)

dagwieers:
I will perform some tests later tonight so we have real metrics to go by. While not 100% accurate it may give some insights on the grand order.

lib/inputstreamhelper/widevine/arm.py
```diff
@@ -162,7 +162,8 @@ def install_widevine_arm(backup_path):  # pylint: disable=too-many-statements
         log(4, 'We could not find an ARM device in the Chrome OS recovery.conf')
         ok_dialog(localize(30004), localize(30005))
         return ''
-    required_diskspace = int(arm_device['filesize']) + int(arm_device['zipfilesize'])
+    # Estimated required disk space: take into account 30 MiB filesystem overhead
```
dagwieers, line 165: I would state:
```python
     # Estimated required disk space: takes into account an extra 30 MiB buffer (incl. a 5MiB EULA)
```

dagwieers:
Here it goes, first in kiB, then in MiB.
### Original filesystem
```
Filesystem           1K-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2        29665087  14479195  15181796  49% /storage
Filesystem           1M-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2           28970     14139     14825  49% /storage
```
### After removing Widevine CDM library *(no change)*
```
Filesystem           1K-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2        29665087  14479196  15181795  49% /storage
Filesystem           1M-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2           28970     14139     14825  49% /storage
```
### After downloading EULA *(-11 MiB ?!?)*
```
Filesystem           1K-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2        29665087  14467541  15193450  49% /storage
Filesystem           1M-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2           28970     14128     14837  49% /storage
```
### It reports that I need *3.1GiB* of free disk space to continue.
### After downloading the ChromeOS recovery image *(+972MiB)*
```
Filesystem           1K-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2        29665087  15462571  14198420  52% /storage
Filesystem           1M-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2           28970     15100     13865  52% /storage
```
### After extraction of the image *(+2220MiB)*
```
Filesystem           1K-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2        29665087  17735860  11925131  60% /storage
Filesystem           1M-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2           28970     17320     11646  60% /storage
```
### When finished *(-3188MiB)*
```
Filesystem           1K-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2        29665087  14471159  15189832  49% /storage
Filesystem           1M-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2           28970     14132     14834  49% /storage
```

Notable things:
- I dumped sizes every 1 second, which is not very good (1 sec can make a difference of about 20MiB)
- After downloading the EULA, I saved 11MiB of disk space (probably some clean up happened?)
- I ended up with 7MiB less disk space use after the complete process
- In total it needed 2220MiB + 972MiB = 3192MiB or 2542410kiB + 995030kiB = 3537440kiB
- According to the image used (BOB) we need 2271240192b + 1101940875b = 3373181067b = 3294122kiB = 3216MiB = 3.141GiB

So the granularity is a problem. I would need to add OK-dialogs to get proper numbers, so I can sync to disk after download and after compress before removal.

dagwieers:
Let's do this one more time, now wait ok-dialogs during each step.

#### Original filesystem
```
Filesystem           1M-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2           28970     14132     14834  49% /storage
Filesystem           1K-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2        29665087  14471163  15189828  49% /storage
```
#### After downloading EULA *(+4MiB)*
```
Filesystem           1K-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2        29665087  14475334  15185657  49% /storage
Filesystem           1M-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2           28970     14136     14830  49% /storage
```
#### It reports that I need *3.1GiB* of free disk space to continue.
#### After downloading the ChromeOS recovery image *(+970MiB)*
```
Filesystem           1K-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2        29665087  15468700  14192291  52% /storage
Filesystem           1M-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2           28970     15106     13860  52% /storage
```
#### After uncompressing the image *(+2214MiB)*
```
kodi01:~ # df /dev/mmcblk0p2; df -m /dev/mmcblk0p2
Filesystem           1K-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2        29665087  17735861  11925130  60% /storage
Filesystem           1M-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2           28970     17320     11646  60% /storage
```
#### On disk this is stored in temp/
```
kodi01:~ # ls -l .kodi/userdata/addon_data/script.module.inputstreamhelper/temp/; \
> ls -lh .kodi/userdata/addon_data/script.module.inputstreamhelper/temp/
total 3256323
-rw-r--r--    1 root     root       4270381 Apr 15 00:41 4.10.1582.2-mac-x64.zip
-rw-r--r--    1 root     root     2312994816 Apr 15 00:48 chromeos_12739.111.1_bob_recovery_stable-channel_mp.bin
-rw-r--r--    1 root     root     1017204780 Apr 15 00:43 chromeos_12739.111.1_bob_recovery_stable-channel_mp.bin.zip
total 3G     
-rw-r--r--    1 root     root        4.1M Apr 15 00:41 4.10.1582.2-mac-x64.zip
-rw-r--r--    1 root     root        2.2G Apr 15 00:48 chromeos_12739.111.1_bob_recovery_stable-channel_mp.bin
-rw-r--r--    1 root     root      970.1M Apr 15 00:43 chromeos_12739.111.1_bob_recovery_stable-channel_mp.bin.zip
```
#### When finished *(-3188MiB)*
```
kodi01:~ # df /dev/mmcblk0p2; df -m /dev/mmcblk0p2
Filesystem           1K-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2        29665087  14471163  15189828  49% /storage
Filesystem           1M-blocks      Used Available Use% Mounted on
/dev/mmcblk0p2           28970     14132     14834  49% /storage
```

Notable things:
- The EULA is downloaded before the calculation so it does not affect the comparison !
- In total it needed 2214MiB + 970MiB = 3184MiB = 3.109GiB
- According to the image (BOB) used we need 2271240192b + 1101940875b = 3373181067b = 3294122kiB = 3216MiB = 3.141GiB
- Looking at the files on disk compared to the reported sizes from recovery.conf, there is a 40MB discrepancy already. We request to have 40MiB more than we actually need.
  - The ZIP file is reported to be 1101940875 in size, but is actually only 1017204780 on disk (a difference of 80.8MiB)
  - The image is reported to be 2271240192 in size, but is actually 2312994816 on disk (a difference of 39.820MiB)
  - *The difference in results with the numbers above are due to rounding errors (as I calculated from the MiB numbers)*
- So if we add another 30MiB we have in effect a 70MiB buffer...

#### Problems
- The progress dialog shows 2x the running time, that's a bug
- After doing this manual update, the info dialog reports I last updated on 2020-04-04, while it should state 2020-04-15 (just now)!
  - And I don't see this date reflected on disk anywhere:
```
kodi01:~ # ls -lh .kodi/userdata/addon_data/script.module.inputstreamhelper/backup/* .kodi/cdm/
.kodi/cdm/:
-rwxr--r--    2 root     root        7.4M Apr 14 23:47 libwidevinecdm.so
-rw-r--r--    2 root     root      163.5K Apr 14 23:47 recovery.conf.json
.kodi/userdata/addon_data/script.module.inputstreamhelper/backup/12607.58.0:
-rwxr--r--    1 root     root        7.4M Jan 26 00:05 libwidevinecdm.so
-rw-r--r--    1 root     root      157.5K Jan 26 00:05 recovery.conf.json
.kodi/userdata/addon_data/script.module.inputstreamhelper/backup/12607.82.0:
-rwxr--r--    1 root     root        7.4M Jan 27 21:33 libwidevinecdm.so
-rw-r--r--    1 root     root      157.6K Jan 27 21:33 recovery.conf.json
.kodi/userdata/addon_data/script.module.inputstreamhelper/backup/12739.111.1:
-rwxr--r--    2 root     root        7.4M Apr 14 23:47 libwidevinecdm.so
-rw-r--r--    2 root     root      163.5K Apr 14 23:47 recovery.conf.json
```
  - Even when running it a second time I don't see the last_updated date change.
- The Cancel button during extraction does not work. Maybe that is a good thing ;-)
- The Restore dialog should give an indication that the version listed is the ChromeOS version

dagwieers:
The trend is clear, newer images now need 3.1GiB, whereas before it was 2.7GiB and before that 2.1GiB :-(

horstle:
If you manually install via the settings page, you have to hit OK and reopen it before looking at the info dialog, since last_update (and any other setting change) is only saved then.

dagwieers:
@horstle Sure but according to file timestamps there was nothing updated on 2020-04-04. There is no backup with that date either. The **last_update** value reflects when it was last checked, not when it was last updated, which is what we report. So if we want to keep both values, we probably should keep both values.

mediaminister:
> The progress dialog shows 2x the running time, that's a bug

I fixed this in https://github.com/emilsvennesson/script.module.inputstreamhelper/pull/316

@dagwieers You can close this, you have proven it's very unlikely this PR fixes something.

dagwieers:
Ok. Although I don't object to adding a small buffer. As I don't know if the discrepancy is always to our advantage as it is here for BOB. The next time another image is selected we may be in a disadvantage and this may have been the cause of past issue??

I think it is strange that the reported sizes in recovery.conf do not correspond to the actual files that are downloaded.

horstle:
> What's more the settings on disk has not been updated either:
> ...
> It still reflects 2020-04-04

You probably hit Cancel instead of OK, when closing the settings page, in which case it won't save the setting. For me it works if I press OK and reopen.
Anyways adding a last_check and differentiating between the two would make sense.

dagwieers:
> You probably hit Cancel instead of OK, when closing the settings page, in which case it won't save the setting. For me it works if I press OK and reopen.

No I did not, and can confirm this on two different systems. Also hitting OK shouldn't be required because it is not being updated through the settings anyway, it is a script that is being called.

I don't think **last_checked** is updated when manually updating Widevine. And may be updated, even when no Widevine was installed. So a mismatch between what it is used for, and how it is being used for reporting.

horstle:
> No I did not, and can confirm this on two different systems.

That's really strange. I cannot reproduce this.

> Also hitting OK shouldn't be required because it is not being updated through the settings anyway, it is a script that is being called.

Well, the script is called through the settings, so in a way... but this isn't something you should discuss with me anyways, I can't change it.

> I don't think **last_checked** is updated when manually updating Widevine. 

It's called __last_update__ and it wasn't updated until #290 was merged.

> And may be updated, even when no Widevine was installed. So a mismatch between what it is used for, and how it is being used for reporting.

The inconsistent usage is why I proposed:
> Anyways adding a last_check and differentiating between the two would make sense.

PR #317, opened 2020-04-16
Fix overlooked call to log()

horstle:
While rebasing something I'm working on I saw this. Change to a different level if you prefer.


emilsvennesson:
@horstle I have sent you a collaborator invite. It's long overdue seeing as you have contributed a lot! :-)

horstle:
Thanks a lot.

PR #322, opened 2020-04-20
Add cached property functions

mediaminister:
After https://github.com/emilsvennesson/script.module.inputstreamhelper/commit/830cfb566133e8e925722bdafe623b5b65ecc420 was merged, unit tests didn't fully execute anymore because they used cached properties for arch() and system_os() information.

This pull request adds get/set/delete wrapper functions for the cached properties and deletes the cached properties when executing unit tests.

### DIFF
--- a/lib/inputstreamhelper/kodiutils.py
+++ b/lib/inputstreamhelper/kodiutils.py
@@ -19,7 +19,7 @@ class progress_dialog(DialogProgress, object):  # pylint: disable=invalid-name,u
         """Create and show a progress dialog"""
         if kodi_version_major() < 19:
             lines = message.split('\n', 2)
-            line1, line2, line3 = (lines + [None] * (3-len(lines)))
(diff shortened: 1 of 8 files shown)


dagwieers:
Yes, so my caching indeed caused issues for the testing framework, but I don't like to add the other interfaces as we do here. The caching is meant to be used for situations where this delete should not be needed. So I would not expose a complete interface for this and keep it as it was.

The only thing needed is a del(arch.cached) in the unit tests IMO.

mediaminister:
Okay, that's also possible.

dagwieers:
Sorry I didn't act on this sooner, but I have a lot of work lately, both with work, privately and the volunteering work for the school (going digital).

dagwieers:
Thanks for picking this up, because I hadn't seen the issue myself during testing. Probably only half awake :-(

mediaminister:
No problem, I'm happy that I could fix it.

PR #323, opened 2020-04-20
Delete arch.cached property in unit tests

mediaminister:
After https://github.com/emilsvennesson/script.module.inputstreamhelper/commit/830cfb566133e8e925722bdafe623b5b65ecc420 was merged, unit tests didn't fully execute anymore because they used the cached property in `arch()`.

Deleting the cached property when executing unit tests solves this bug.

### DIFF
--- a/tests/test_ishelper_android_arm.py
+++ b/tests/test_ishelper_android_arm.py
@@ -8,6 +8,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 import unittest
 import platform
 import inputstreamhelper
+from test_utils import delete_cached
 
 xbmc = __import__('xbmc')
 xbmcaddon = __import__('xbmcaddon')
@@ -16,6 +17,9 @@ xbmcvfs = __import__('xbmcvfs')
 
 
 class AndroidARMTests(unittest.TestCase):
+
+    def setUp(self):
+        delete_cached()
 
     def test_check_inputstream_mpd(self):
         inputstreamhelper.system_os = lambda: 'Android'
--- a/tests/test_ishelper_macos_x64.py
+++ b/tests/test_ishelper_macos_x64.py
@@ -8,6 +8,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 import unittest
 import platform
 import inputstreamhelper
+from test_utils import delete_cached
 
 xbmc = __import__('xbmc')
 xbmcaddon = __import__('xbmcaddon')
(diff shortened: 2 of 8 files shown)


horstle:
Instead of adding those two lines multiple times, I think it would make sense to make a function out of it (maybe in a new test_utils.py) and adding that. If more functions get caches in the future, we would only have to add a line to remove them at one place.

mediaminister:
Okay, I made it little more future-proof.

tests/test_ishelper_android_arm.py
```diff
@@ -18,6 +19,7 @@
 class AndroidARMTests(unittest.TestCase):
 
     def test_check_inputstream_mpd(self):
+        delete_cached()
```
dagwieers, line 22: This only needs to happen once per class either in the `setUp()` or in the `tearDown()`method as it is fine to keep it cached within the class. But this will work as well :-)

dagwieers:
@horstle We want to test the caching as well, so we shouldn't be deleting all the caches all the time. The only problem here is that caching `arch()` is problematic as we fake different architectures within the same framework. Which is impossible to happen in real life. For other cached values this may not be problematic, unless maybe they are also architecture-dependent.

But caching values only really makes sense to avoid expensive calls. And we don't have that many. I don't even know if `arch()` is really expensive or whether it gets called that often. But it is easy to do.

tests/test_ishelper_android_arm.py
mediaminister: Didn't know that, even better.

dagwieers: This is very specific to **unittest** and we already do this fro VRT NU to test our code using different settings so we get coverage for bits of code hidden behind global (e.g. proxy) or add-on settings.

tests/test_ishelper_macos_x64.py
```diff
@@ -41,6 +45,7 @@ def test_check_inputstream_rtmp(self):
         self.assertTrue(is_installed, True)
 
     def test_check_inputstream_disabled(self):
+        delete_cached()
```
dagwieers, line 48: This can be removed too.

tests/test_ishelper_windows_x64.py
```diff
@@ -41,6 +45,7 @@ def test_check_inputstream_rtmp(self):
         self.assertTrue(is_installed, True)
 
     def test_check_inputstream_disabled(self):
+        delete_cached()
```
dagwieers, line 48: This can be removed too.

dagwieers:
So personally I am not that a fan of doing this single cache cleanup from a function (because you don't see what is done exactly, and cannot see why) and it is only a single call, but that's just a nitpick. Functionally it is one and the same.

PR #326, opened 2020-04-26
Use temp path with unicode chars for testing

mediaminister:
In the past we never tested temp paths containing unicode characters and unicode bugs were left undetected. We could only fix these bugs after a user created an issue.

This pull request fixes a couple of unicode bugs and adds unicode chars to the tests temp_path in addon_settings.json.

### DIFF
--- a/tests/xbmcaddon.py
+++ b/tests/xbmcaddon.py
@@ -6,7 +6,6 @@
 # pylint: disable=invalid-name
 
 from __future__ import absolute_import, division, print_function, unicode_literals
-from xbmc import getLocalizedString
 from xbmcextra import ADDON_ID, ADDON_INFO, addon_settings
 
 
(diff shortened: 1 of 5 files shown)


tests/xbmcaddon.py
```diff
@@ -6,7 +6,6 @@
 # pylint: disable=invalid-name
 
 from __future__ import absolute_import, division, print_function, unicode_literals
-from xbmc import getLocalizedString
```
horstle, line 9: I'd probably have kept this import here, maybe using import ... as ..., but I'm fine with this, too.
Once restructuring is done, I think we can move imports back to the top, then we can have another thought about this.

mediaminister: I did this to avoid a circular import in the CLI test (`make test-run`)

horstle: Ok, in the long run this should probably be solved differently, but that shouldn't affect this PR. I'll merge this in ~2h if there are no objections.

PR #327, opened 2020-04-26
Make chromeos_offset work on bytestream

horstle:
With this change the chromeos offset can be calculated from a bytestream of the .bin.
One advantage is this works without the help of `fdisk` or `parted`, so no shell command needs to be called. But this also brings us one step closer to #72, since this already also works when executed on `ZipFile('chromeos....bin.zip', 'r').open('chromeos....bin', 'r')`.

### DIFF
--- a/lib/inputstreamhelper/widevine/arm.py
+++ b/lib/inputstreamhelper/widevine/arm.py
@@ -5,6 +5,7 @@
 from __future__ import absolute_import, division, unicode_literals
 import os
 from time import time
+from struct import unpack, calcsize
 
 from .. import config
 from ..kodiutils import browsesingle, copy, exists, localize, log, mkdir, ok_dialog, open_file, progress_dialog, yesno_dialog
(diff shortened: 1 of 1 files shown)


lib/inputstreamhelper/widevine/arm.py
```diff
@@ -21,25 +22,50 @@ def mnt_path():
...
+        part_name = part_name.decode('utf-16').strip('\x00')
+        if part_name == 'ROOT-A':
+            offset = first_lba * lba_size
+            break
+
+    if not offset:
+        log(4, 'Failed to calculate losetup offset.')
+        return '0'
+
+    return str(offset)
```
mediaminister, line 68: I thinks it's more logical to return the integer and convert to string when needed in another process.

horstle: You're right. Seems more logical to me, too.

horstle:
It probably makes more sense to directly merge #348 when that's ready, so I'll close this now.

Issue #333, opened 2020-05-01
'ascii' codec can't decode byte 0xc5 in position 29


Cumol:
I am running into the same error as in https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/92

It happens during the extraction of widevine

I am on Kodi 18.6 Git_leia_pi4_18.6-Leia
Inputstream Helper 0.4.6
InputStream adaptive 2.4.4.1

```
2020-05-01 07:18:42.860 T:2573202288   ERROR: [plugin.video.netflix (1)] Traceback (most recent call last):
                                              File "/storage/.kodi/addons/plugin.video.netflix/resources/lib/run_addon.py", line 207, in run
                                                route([part for part in g.PATH.split('/') if part])
                                              File "/storage/.kodi/addons/plugin.video.netflix/resources/lib/run_addon.py", line 38, in lazy_login_wrapper
                                                return func(*args, **kwargs)
                                              File "/storage/.kodi/addons/plugin.video.netflix/resources/lib/run_addon.py", line 62, in route
                                                play(videoid=pathitems[1:])
                                              File "/storage/.kodi/addons/plugin.video.netflix/resources/lib/common/videoid.py", line 295, in wrapper
                                                return func(*args, **kwargs)
                                              File "/storage/.kodi/addons/plugin.video.netflix/resources/lib/common/logging.py", line 143, in timing_wrapper
                                                return func(*args, **kwargs)
                                              File "/storage/.kodi/addons/plugin.video.netflix/resources/lib/navigation/player.py", line 82, in play
                                                list_item = get_inputstream_listitem(videoid)
                                              File "/storage/.kodi/addons/plugin.video.netflix/resources/lib/navigation/player.py", line 163, in get_inputstream_listitem
                                                if not is_helper.check_inputstream():
                                              File "/storage/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/__init__.py", line 407, in check_inputstream
                                                return self._check_drm()
                                              File "/storage/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/__init__.py", line 362, in _check_drm
                                                return self.install_widevine()
                                              File "/storage/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/__init__.py", line 33, in clean_before_after
                                                result = func(self, *args, **kwargs)
[...]

PR #334, opened 2020-05-01
Support unicode chars in paths

mediaminister:
This fixes https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/333

### DIFF
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -7,12 +7,13 @@ import os
 
 from . import config
 from .kodiutils import (addon_version, delete, exists, get_proxies, get_setting, get_setting_bool, get_setting_float, get_setting_int, jsonrpc,
-                        kodi_to_ascii, kodi_version, localize, log, notification, ok_dialog, progress_dialog, select_dialog,
+                        kodi_to_ascii, kodi_version, listdir, localize, log, notification, ok_dialog, progress_dialog, select_dialog,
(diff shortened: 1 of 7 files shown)


horstle:
I think `mnt_path()` should return `to_unicode(mount_path)`, then it should actually fix the issue.

horstle:
So the paths that I'd add to_unicode to are:
backup_path
mount_path
temp_path

mediaminister:
I'll take another look at the other path functions, but the root cause is `xbmc.translatePath` in Kodi 18.x (Python 2.7) which needs a bytestring as input and returns a bytestring as output.

So, I ensured that the `translate_path` wrapper function does the necessary conversions in Python 2.7 with `to_unicode` and `from_unicode`.

horstle:
Right, I forgot you added a test for a non-ascii temp_path already.
But `translate_path` already had a `to_unicode` before, so the output should already have been unicode, yet still `mount_path` is a bytestring.

horstle:
Are you sure the output of `xbmc.translatePath` is a "normal" bytestring and not something kodi specific, which wouldn't be recognized by `isinstance(text, bytes)`?

mediaminister:
> Are you sure the output of xbmc.translatePath is a "normal" bytestring
Yes, it's a bytestring in Kodi 18.x Python2

Testing unicode was unreliable because `unicode_literals` was imported in the tests. I'm trying to fix this now. Thinking how to mimic Kodi 18 and 19 behaviour with `addon_settings` in xbmcextra.py

Issue #305, opened 2020-04-06
Download times out


Mintux92:
Good morning,
I am installing neflix on libreelec. Netflix installs without problems, when I try to log in it asks me to install the various addons but when it tries chrome os at some point it stops and in the logs I find what I attach. How could I solve it?

KODI 18.2
ARCHITECTURE: RPi 3 B+
Repository: Castagnait Repository 1.1.1

log:
```
2020-04-05 20:49:55.974 T:1545548656 ERROR: [plugin.video.netflix (1)] Traceback (most recent call last):
File "/storage/.kodi/addons/plugin.video.netflix/resources/lib/run_addon.py", line 182, in run
route([part for part in g.PATH.split('/') if part])
File "/storage/.kodi/addons/plugin.video.netflix/resources/lib/run_addon.py", line 33, in lazy_login_wrapper
return func(*args, **kwargs)
File "/storage/.kodi/addons/plugin.video.netflix/resources/lib/run_addon.py", line 57, in route
play(videoid=pathitems[1:])
File "/storage/.kodi/addons/plugin.video.netflix/resources/lib/common/videoid.py", line 294, in wrapper
return func(*args, **kwargs)
File "/storage/.kodi/addons/plugin.video.netflix/resources/lib/common/logging.py", line 133, in timing_wrapper
return func(*args, **kwargs)
File "/storage/.kodi/addons/plugin.video.netflix/resources/lib/navigation/player.py", line 72, in play
list_item = get_inputstream_listitem(videoid)
File "/storage/.kodi/addons/plugin.video.netflix/resources/lib/navigation/player.py", line 144, in get_inputstream_listitem
if not is_helper.check_inputstream():
File "/storage/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/init.py", line 1053, in check_inputstream
return self._check_drm()
File "/storage/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/init.py", line 1008, in _check_drm
return self.install_widevine()
File "/storage/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/init.py", line 776, in install_widevine
return self._install_widevine_arm()
File "/storage/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/init.py", line 703, in _install_widevine_arm
[...]

PR #339, opened 2020-05-01
Track real progress of unzip in install_widevine_arm

### LINKED ISSUE #325: Widevine cdm extraction stuck
I installed an add-on which requires widevine-cdm on LibreElec on my Raspberry Pi 4, it correctly opens the EULA window and warns you it'll have to download a 3.1G ChromeOS image. It correctly downloads it, then starts extracting and the funny thing is that the more the progress bar goes on the slower that progress becomes.

I tried a few times for an hour or so, tonight I tried leaving it working for around 12h and this morning it was *exactly* where I left it yesterday night. I checked my [kod.log](https://gist.github.com/giovariot/30859d41e2b5c222b36ad6c109224f9f) but couldn't find anything wrong.

I also read this [comment](https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/292#issuecomment-596257643) in issue #292, I tried to run it but being a complete and utter noob this is what I get:
``Traceback (most recent call last):
  File "load-chromium-widevine.py", line 137, in <module>
    os.remove("recovery.conf")
FileNotFoundError: [Errno 2] No such file or directory: 'recovery.conf'``

Does anyone know where I can get the right recovery.conf? Is it the one suggested [here](https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/26#issuecomment-392329000) on #26 ?

Is there any way I can get widevine to install? I also tried copying libwidevinecdm.so to .kodi/cdm/. but I'm told the system needs to update widevine cdm and sends me back to downloading the chromeos image. I know that's from not adding the correct configuration into the json file, but I really don't know what to do.

Thanks in advance

horstle:
This changes the way the ChromeOS image is extracted and tracks the progress. It most probably fixes #325 .

With this change, the `unzip()` function isn't used anywhere at all, but I wasn't sure, if it should be removed.

### DIFF
--- a/resources/language/resource.language.de_de/strings.po
+++ b/resources/language/resource.language.de_de/strings.po
@@ -190,8 +190,8 @@ msgid "Uncompressing image..."
 msgstr "Entpacke das Image..."
 
 msgctxt "#30046"
-msgid "This may take several minutes. ({mins:d}:{secs:02d})"
-msgstr "Dies kann einige Minuten in Anspruch nehmen. ({mins:d}:{secs:02d})"
+msgid "This may take several minutes."
+msgstr "Dies kann einige Minuten in Anspruch nehmen."
 
 msgctxt "#30047"
 msgid "[I]Please do not interrupt this process.[/I]"
@@ -236,6 +236,10 @@ msgstr "Keine Backups gefunden!"
 msgctxt "#30057"
 msgid "Choose a backup to restore"
 msgstr "Wählen Sie ein Backup zum Wiederherstellen aus."
+
+msgctxt "#30058"
+msgid "Time remaining: {mins:d}:{secs:02d}"
+msgstr ""
 
 
 ### INFORMATION DIALOG
--- a/lib/inputstreamhelper/utils.py
+++ b/lib/inputstreamhelper/utils.py
@@ -4,6 +4,7 @@
 
 from __future__ import absolute_import, division, unicode_literals
 import os
+from time import time
 
 from . import config
 from .kodiutils import copy, delete, exists, get_setting, localize, log, mkdirs, ok_dialog, progress_dialog, set_setting, stat_file, translate_path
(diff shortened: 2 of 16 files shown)


resources/language/resource.language.de_de/strings.po
```diff
@@ -190,8 +190,8 @@ msgid "Uncompressing image..."
 msgstr "Entpacke das Image..."
 
 msgctxt "#30046"
-msgid "This may take several minutes. ({mins:d}:{secs:02d})"
-msgstr "Dies kann einige Minuten in Anspruch nehmen. ({mins:d}:{secs:02d})"
+msgid "This may take several minutes."
```
dagwieers, line 193: The reason we show time here is to make sure the user knows things are progressing.
Otherwise he may think things are stalling and may cancel (which is worse as the thread may still be continuing).

So I prefer we update something every second just in case.

dagwieers: Maybe the percentage as floating point?

horstle: In my tests on an RPi3 the updates of the progress bar seemed fast enough to me. I think even on slower devices this should be fine.

horstle: During download the progress bar might increase even less frequently, depending on your network connection.

dagwieers: Correct, during download we have the same problem.

And the main problem is that we started off with a 600MB file, but now newer ZIP files are 2x and even 2.5x that size. And given that older RPi devices have 1GB or less memory (resulting in smaller file caches) it will take proportionally longer than that if file sizes increase.

So I'd rather have progress be updated with a counter or time advancing, also during download. 
(In fact we could give an indication of the remaining time based on current performance)

horstle: > (In fact we could give an indication of the remaining time based on current performance)

That's a good idea. I implemented a rather basic version.

horstle:
Well, seems like the checks just triggered #305 . :laughing:

PR #343, opened 2020-05-03
Fix paths

mediaminister:


lib/inputstreamhelper/widevine/arm.py
```diff
@@ -217,12 +217,12 @@ def install_widevine_arm(backup_path):  # pylint: disable=too-many-statements
...
             from zipfile import ZipFile
 
-            with ZipFile(store('download_path')) as zip_obj:
+            with ZipFile(compat_path(store('download_path'))) as zip_obj:
                 bin_size = zip_obj.getinfo(bin_filename).file_size
                 chunksize = 1024**2
 
                 with zip_obj.open(bin_filename) as member:
-                    with open_file(bin_path, 'wb') as bin_file:
+                    with open(bin_path, 'wb') as bin_file:
```
horstle, line 225: Why did you change this?

mediaminister: I got an error with `open_file` reading binary files on Kodi 19 on Windows. I don't think the `wb` option fully works in `xbmcvfs`. Just to be sure, I removed all open_file functions that needed binary reading/writing

PR #319, opened 2020-04-16
Add udisks, p7zip.

horstle:
This adds the option to use udisks or p7zip to avoid running as root or with sudo.
Also, it changes the mounting/loop workflow, resulting in a simpler cleanup.
Then there are some additional small improvements.

This fixes #246.

Issue #344, opened 2020-05-03
Inconsistent last_update


horstle:
I think we should have another thought about `last_update`, which iirc should more appropriately be called `last_checked` or something. Maybe we could also add a real `last_updated`, which would reflect when a new version was installed. Not sure whether that is needed and how to handle backups/rollbacks though.

mediaminister:
I think we should keep this as simple as possible. We only need the timestamp of the previous update check. This can be called `prev_update_check`. 

FYI, the current implementation doesn't work as expected, I got the following remark from a user a couple of weeks ago:

> I had an update of Widevine yesterday, and it occurred to me that I had to approve the update quite a few times (3 or 4 times if I remember correctly). One messages just showed that I had to have a few command line tools, but I assume it can be detected fine if I have these on my system so there is no need to show this message.
> 6:28
> I think two messages would be fine. One to approve the update, and one to accept the EULA.

dagwieers:
@horstle This is what I would prefer. I'd like to know when it was last checked, and I would like to know when it was last updated.

If we roll back, we should update last_updated as well IMO.

dagwieers:
Let's fix this for v0.5.0 as well.

horstle:
One part of this is pretty much done. Now we have to decide what to do about the remaining usage of last_update.

Some ideas:
* Remove it. Instead, show last check in info dialog.
* Keep it tied to the widevine version (i.e. when that version was installed), e.g. in the manifest.json. In this case, should it be changed when the same version is installed again manually?
* Set to current time whenever a different version is installed, whether it's a new one or from backup.

dagwieers:
Thanks for this!

I would use the current time every time a change is made. This reflects when it was updated, and not "what was updated" or "when it was downloaded first".

horstle:
So any time the content of the cdm directory is changed, the time should be updated?
The only functions that change it are `install_cdm_from_backup` and `remove_widevine`, so we'd only need to set that setting there.

dagwieers:
Everytime the CDM is changed, that value should be update IMO.
We could call `last_update` → `last_changed`, and maybe call `last_check` → `last_checked`.

What do others think?

mediaminister:
The information displayed in the GUI must be unambiguous to the user. Currently, this is not the case.

I think the GUI should show this:
- Widevine not found --> don't display a date
- Widevine version found --> display installation date

[...]

PR #348, opened 2020-05-05
Access the zipped Chrome OS image directly

### LINKED ISSUE #72: Access the zipped image directly to extract CDM

horstle:
Okay, as promised in https://github.com/emilsvennesson/script.module.inputstreamhelper/pull/319#issuecomment-623191660, here is another PR about installing Widevine on ARM.

It accesses the zipped image directly, and extracts the lib from it. On an RPi3 this takes around half the time it takes to extract the whole image and (obviously) only uses about one third of disk space.

I tested this on an RPi3 and it seems to work. The reason it's still a Draft PR is because the progress dialog needs improvement and there are some messages not yet localized (mainly in the progress dialog).
But you're welcome to test it already, the only changes I have planned are the two points mentioned above and they shouldn't affect the actual code.

This is based on #327 and (imho) supersedes #319 . It resolves #72 .

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,9 +1,10 @@
 [MESSAGES CONTROL]
 disable=
     bad-option-value,
-    duplicate-code,
-    import-outside-toplevel,
-    line-too-long,
+    duplicate-code,  # Integration tests tend to test similar things
+    fixme,  # Allow FIXMEs in code is a useful practice
+    import-outside-toplevel,  # Done on purpose to reduce overhead at startup
+    line-too-long,  # Flake8 is used for testing line-length
     old-style-class,
     too-many-arguments,
     too-many-branches,
--- a/lib/inputstreamhelper/utils.py
+++ b/lib/inputstreamhelper/utils.py
@@ -121,10 +121,14 @@ def http_download(url, message=None, checksum=None, hash_alg='sha1', dl_size=Non
             )
 
     if checksum and not calc_checksum.hexdigest() == checksum:
+        progress.close()
+        req.close()
         log(4, 'Download failed, checksums do not match!')
         return False
 
     if dl_size and not stat_file(download_path).st_size() == dl_size:
+        progress.close()
+        req.close()
         log(4, 'Download failed, filesize does not match!')
         return False
 
--- a/lib/inputstreamhelper/widevine/arm.py
+++ b/lib/inputstreamhelper/widevine/arm.py
@@ -4,13 +4,14 @@
 
 from __future__ import absolute_import, division, unicode_literals
 import os
+import json
 from time import time
-from struct import unpack, calcsize
 
 from .. import config
 from ..kodiutils import browsesingle, copy, exists, localize, log, mkdir, ok_dialog, open_file, progress_dialog, yesno_dialog
 from ..utils import cmd_exists, diskspace, http_download, http_get, run_cmd, sizeof_fmt, store, system_os, temp_path, update_temp_path
 from ..unicodes import compat_path, to_unicode
+from .arm_chromeos import ChromeOSImage
 
 
 def mnt_path():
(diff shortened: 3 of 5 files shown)


dagwieers:
I wanted to fix the checks, but I have no permission on your branch.
```diff
diff --git .pylintrc .pylintrc
index 89bedcf..1a60cdd 100644
--- .pylintrc
+++ .pylintrc
@@ -1,9 +1,10 @@
 [MESSAGES CONTROL]
 disable=
     bad-option-value,
-    duplicate-code,
-    import-outside-toplevel,
-    line-too-long,
+    duplicate-code,  # Integration tests tend to test similar things
+    fixme,  # Allow FIXMEs in code is a useful practice
+    import-outside-toplevel,  # Done on purpose to reduce overhead at startup
+    line-too-long,  # Flake8 is used for testing line-length
     old-style-class,
     too-many-arguments,
     too-many-branches,
diff --git lib/inputstreamhelper/widevine/arm.py lib/inputstreamhelper/widevine/arm.py
index 45048cf..a973d1a 100644
--- lib/inputstreamhelper/widevine/arm.py
+++ lib/inputstreamhelper/widevine/arm.py
@@ -190,7 +190,7 @@ def install_widevine_arm(backup_path):
     return False
 
 
-def install_wv_arm_legacy(backup_path):  # pylint: disable=too-many-statements
+def install_wv_arm_legacy(backup_path):
     """Legacy method to install Widevine CDM on ARM-based architectures."""
     root_cmds = ['mount', 'umount', 'losetup', 'modprobe']
     devices = chromeos_config()
```
I think it is convenient/practical to fix small issues like this directly, rather than have to point them out and have someone else fix them.

dagwieers:
This is looking really good. I tested it briefly and I have only 2 additional minor items:
- The download progress dialog is nice, however I would probably only update remaining time every 5 seconds, but I would add a download rate (e.g. 568 kB/sec) which does update every second (so we have changes in the dialog, and it will show 0 kB/sec when the download stalls)
- What I find disturbing visually is that the download of the license is only briefly showing a progress dialog. IMO that dialog is useless as the download usually will only take a few seconds (in my case it seems sub-second). If the plan is online streaming, we can expect people to have a big pipe anyway.

Thanks for this! I spend quite some time looking at implementing something like this, but based on existing libraries/tools. If I have one concern, then it is that it may be more prone to changes to the image. The upside is that if such changes are coming, we can easily fix those (whereas with the old implementation that would be much harder to come by).

We hope that with the Google LaCrOS plans an easier ARM download (and who knows ARM64) will become available. I frequently test for new widevine ZIP files but have not detected any yet.

horstle:
> I think it is convenient/practical to fix small issues like this directly, rather than have to point them out and have someone else fix them.

I'm currently working on and force-pushing to that branch, so I might have accidentally overwritten your changes, but I'll give you write access soon. I added your changes now.

>     * The download progress dialog is nice, however I would probably only update remaining time every 5 seconds or delay showing the ETA with 10 seconds (given the great variance, especially at the start), but I would add a download rate (e.g. 568 kB/sec) so we have changes in the dialog, and it will show download stalls (i.e. 0 kB/sec)
> 
>     * What I find disturbing visually is that the download of the license is only briefly showing a progress dialog. IMO that dialog is useless as the download usually will only take a few seconds (in my case it seems sub-second). If the plan is online streaming, we can expect people to have a big pipe anyway.

There is definitely room for improvement here, however that's not related to this PR (this was changed in #339 ), so I think this should get it's own issue.

> Thanks for this! I spend quite some time looking at implementing something like this, but based on existing libraries/tools. If I have one concern, then it is that it may be more prone to changes to the image. The upside is that if such changes are coming, we can easily fix those (whereas with the old implementation that would be much harder to come by).

Do you have any specific change in mind, which would break this? The filesystem changing or the partition name not being 'ROOT-A' anymore are changes that I thought of, but I think they are rather improbable (for now, might change once ARM64 is there).

horstle:
Alright, I think it's ready now. On an RPi3 the progress dialog might stay the same for a few seconds, but IMO that's not too long, even if it was longer on slower hardware.

dagwieers:
> Do you have any specific change in mind, which would break this? The filesystem changing or the partition name not being 'ROOT-A' anymore are changes that I thought of, but I think they are rather improbable (for now, might change once ARM64 is there).

I think we make the following assumptions:
- GPT is used
- Partition name is `ROOT-A`
- Filesystem is ext2

But as I said, this is no worse than we had before. In fact, it will be easier to fix in our code than having tools no longer being able to cope with the changes.

I think we need another change I will propose for #344 which is that we also store the platform name (image) used. If we get reports it might be because of a broken/incompatible ZIP image and if this would be a temporary situation, we cannot easily reproduce without this information.

Same for a broken Widevine, it would be nice to find where it was coming from.

dagwieers:
I would still update the progress dialog with more meaningful messages. "Reading the file..." does not bring anything to the table. We might as well keep saying "Extracing Widevine CDM...".

I think we can give a bit more insightful message, like:
- Opening compressed ChromeOS image...
- Identifying partition...
- Reading file system...
- Scanning the file system for Widevine CDM...
- Widevine CDM found, extracting to disk...

For every stage where we may expect trouble in the future it could be useful to update the message so that the user can report "When it failed, the progress bar showed: ...". While we prefer a full log, it helps if the user can easily relate their issue with existing open issues.

dagwieers:
Let's merge this, and improve whatever needs improving. I guess this is the start of the v0.5.0 development cycle. Welcome, and thanks @horstle !

dagwieers:
Wrt. potential future changes to the ChromeOS images, what we could do is run our unit tests for ARM in a **Status** GitHub workflow e.g. every 2 hours just to make sure we have an advance notice when it starts failing. (Even better would be to only test when a new selected image was released...)

PR #350, opened 2020-05-06
Improvements for #348

horstle:
Here are some small improvements to #348 as discussed there, including
* comments where assumptions are made
* a slightly improved progress dialog
* get_bstream is now a static method


mediaminister:
I tested this on LibreELEC 9.2.2 VM with faked `sys_arch = 'arm'` and it works. It's okay to merge this.

Issue #351, opened 2020-05-15
LooseVersion instance has no attribute 'version'


npmstart-pray:
so there I was 15 minutes ago watching netflix, and when I hit Watch when it came up in Up Next, I got a message about WideVine not being installed, and then this error came up, telling me to report it.
Ant idea what went wrong and/or how to fix it?
4GB RasPi 4, current xbian k5.4.38  kodi 18.6 buster current inpustream helper
I'll go digging for the log and post here.

zurrealist:
I'm running into this same issue. I'm on Kodi 18.6, running LibreELEC 9.2.1 on a RasPi 4.
Netflic worked fine yesterday, but today I'm getting an error that "An error occurred while installing Widevine CDM" and to post here. The following message reads "The operation has been cancelled. InputStream Helper has been generated an internal error: LooseVersion instance has no attribute 'version'"
Log should be at:
https://paste.kodi.tv/osewufiyoc

npmstart-pray:
I was hoping for a fix in the documentation, but no such luck. Part of me wonders if it went looking for (or was notified of) a widevine update and failed.

zurrealist:
I did try reinstalling both the Netflix add-on and the InputStream Adaptive add-on with no luck.

horstle:
Anyone experiencing this issue or #352 :

If you could previously watch videos, you can simply disable inputstreamhelper for now. To do that, go to the Addons section, open Inputstream Helper and enable "Disabled". After that your videos should start as before.

horstle:
The problem is, that the recovery.conf contains nothing. Hopefully Google fixes this, soon. But in the long run, we should analyze the "Chrome OS recovery utility", assuming that doesn't use the recovery.conf.

horstle:
Ok, the utility uses https://dl.google.com/dl/edgedl/chromeos/recovery/recovery.json or https://dl.google.com/dl/edgedl/chromeos/recovery/onhub_recovery.json.

The onhub_recovery.json only contains ARKHAM, WHIRLWIND and GALE, which afaik are all incompatible.
The recovery.json is completely empty. I guess some Google internal script which should update recovery.conf and recovery.json failed and created empty files instead.

dagwieers:
[...]

PR #353, opened 2020-05-15
Handle missing widevine config or unknown latest version in update_widevine

horstle:
This relates to #351 and fixes it for those who already have widevine installed.

### DIFF
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -274,13 +274,21 @@ class Helper:
                 return
 
         wv_config = load_widevine_config()
-        latest_version = latest_widevine_version()
-        if 'x86' in arch():
+        if not wv_config:
+            log(3, 'Widevine config missing. Could not determine current version, forcing update.')
+            current_version = '0'
+        elif 'x86' in arch():
             component = 'Widevine CDM'
             current_version = wv_config['version']
         else:
             component = 'Chrome OS'
             current_version = select_best_chromeos_image(wv_config)['version']
+
+        latest_version = latest_widevine_version()
+        if not latest_version:
+            log(3, 'Updating widevine failed. Could not determine latest version.')
+            return
+
         log(0, 'Latest {component} version is {version}', component=component, version=latest_version)
         log(0, 'Current {component} version installed is {version}', component=component, version=current_version)
 


lib/inputstreamhelper/__init__.py
```diff
@@ -274,13 +274,21 @@ def _update_widevine(self):
                 return
 
         wv_config = load_widevine_config()
-        latest_version = latest_widevine_version()
-        if 'x86' in arch():
+        if not wv_config:
+            log(3, 'Widevine config missing. Could not determine current version, forcing update.')
```
dagwieers, line 278: I don't understand the error, if the Widevine config is missing, there is nothing to update to.
It should report and return.

horstle: load_widevine_config reads the config of the currently installed version.

vtolstov:
for users that not have installed cdm this is not helps
in rpi4 i get
We could not find an ARM device in the Chrome OS recovery.conf

rtucker-scs:
Is there a fix for people who do not have widevine installed?

mediaminister:
> Is there a fix for people who do not have widevine installed?

No, this add-on needs the recovery.conf file from Google Chrome OS and this file is currently offline.
Therefore the Chromebook Recovery Utility is also broken, this has an impact on all Chrome OS users worldwide who need to reinstall their Chrome OS device, so hopefully Google will fix this very soon.

To help, you can upvote the issue at the Chromebook community: https://support.google.com/chromebook/thread/46853544

InputStream Helper add-on will work again from the moment Google fixes this problem.

lib/inputstreamhelper/__init__.py
dagwieers: Ah, ok. So this is the case where no Widevine was ever installed through InputStreamHelper?
But how about the user added its own Widevine CDM manually?

Looking at the code now, this is all rather confusing. We get the current_version from "select_best_chromeos_image" ?
Either we should add some comments to indicate why or the function calls need to reflect this better. Not just here, but in various other places as well.

(This is no criticism to anyone, it's just that I haven't looked at this code for a while and I got totally confused now :-))

lib/inputstreamhelper/__init__.py
```diff
@@ -274,13 +274,21 @@ def _update_widevine(self):
...
+        elif 'x86' in arch():
             component = 'Widevine CDM'
             current_version = wv_config['version']
         else:
             component = 'Chrome OS'
             current_version = select_best_chromeos_image(wv_config)['version']
+
+        latest_version = latest_widevine_version()
+        if not latest_version:
+            log(3, 'Updating widevine failed. Could not determine latest version.')
```
dagwieers, line 289: I wouldn't state "Updating widevine failed" because that sounds rather fatal. I would go for something like `Skipping widevine update because latest version could no be determined.` Or even more specific if we know why determining latest version failed.

lib/inputstreamhelper/__init__.py
horstle: If a user manually places the libwidevinecdm.so in the cdm directory, he should have enough expertise to simply disable inputstreamhelper.

For ARM, the whole recovery.conf (with all the devices) is saved in the same place as the libwidevinecdm.so extracted from one of the images in that conf. The current version is determined by running select_best_chromeos_image on said recovery.conf.
I agree, that keeping the whole recovery.conf is unnecessary, but please open a new issue for that, as it's unrelated to this.

lib/inputstreamhelper/__init__.py
horstle: Well, it's just what it is. update_widevine was called, but couldn't update. That's not skipping an update, that's an update failing.

PR #357, opened 2020-05-20
Improvements to http dl and request

### LINKED ISSUE #281: Error occured while installing Widevine CDM
After installing Netflix pluggin i have to install Widevine CDM. Everything was ok (i have to use USB flash disk for download Chrome os recovery image), Ater download unpacking have started and error has occured. 

[kodi.log](https://github.com/emilsvennesson/script.module.inputstreamhelper/files/4270301/kodi.log)
[kodi_crashlog-20200229_133814.log](https://github.com/emilsvennesson/script.module.inputstreamhelper/files/4270303/kodi_crashlog-20200229_133814.log)

### LINKED ISSUE #305: Download times out

horstle:
This includes:
* Report free disk space if after download the filesizes do not match (as discussed https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/281#issuecomment-614294737)
* Retry downloading if it fails, base on code from @mediaminister here https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/305#issuecomment-627511506

This resolves #281 and fixes #305, but there are two points which I'd like to discuss before merging:

* `total_length` is a float, I think it should be int?
* What should be done if the `dl_size` (the zipfilesize of the Chrome OS image extracted from recovery.conf) does not match `total_length`?

### DIFF
--- a/lib/inputstreamhelper/utils.py
+++ b/lib/inputstreamhelper/utils.py
@@ -86,7 +86,11 @@ def http_download(url, message=None, checksum=None, hash_alg='sha1', dl_size=Non
         message = localize(30015, filename=filename)  # Downloading file
 
     download_path = os.path.join(temp_path(), filename)
-    total_length = float(req.info().get('content-length'))
+    total_length = int(req.info().get('content-length'))
+    if dl_size and dl_size != total_length:
+        log(2, 'The given file size does not match the request!')
+        dl_size = total_length  # Otherwise size check at end would fail even if dl succeeded
+
     if background:
         progress = bg_progress_dialog()
     else:
@@ -120,16 +124,17 @@ def http_download(url, message=None, checksum=None, hash_alg='sha1', dl_size=Non
 
             progress.update(percent, prog_message)
 
-    if checksum and not calc_checksum.hexdigest() == checksum:
+    if checksum and calc_checksum.hexdigest() != checksum:
         progress.close()
         req.close()
         log(4, 'Download failed, checksums do not match!')
         return False
 
-    if dl_size and not stat_file(download_path).st_size() == dl_size:
+    if dl_size and stat_file(download_path).st_size() != dl_size:
         progress.close()
         req.close()
-        log(4, 'Download failed, filesize does not match!')
+        free_space = sizeof_fmt(diskspace())
+        log(4, 'Download failed, filesize does not match! Filesystem full? Remaining diskspace in temp: {}.'.format(free_space))
         return False
 
     progress.close()
--- a/lib/inputstreamhelper/utils.py
+++ b/lib/inputstreamhelper/utils.py
@@ -5,9 +5,17 @@
 from __future__ import absolute_import, division, unicode_literals
 import os
 from time import time
+from socket import timeout
+from ssl import SSLError
+
+try:  # Python 3
+    from urllib.error import HTTPError, URLError
+    from urllib.request import Request, urlopen
+except ImportError:  # Python 2
+    from urllib2 import HTTPError, Request, URLError, urlopen
 
 from . import config
-from .kodiutils import bg_progress_dialog, copy, delete, exists, get_setting, localize, log, mkdirs, ok_dialog, progress_dialog, set_setting, stat_file, translate_path
+from .kodiutils import bg_progress_dialog, copy, delete, exists, get_setting, localize, log, mkdirs, progress_dialog, set_setting, stat_file, translate_path, yesno_dialog
 from .unicodes import compat_path, from_unicode, to_unicode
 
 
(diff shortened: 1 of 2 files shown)


mediaminister:
> total_length is a float, I think it should be int?

We need to pass at least one `float` when calculating `percent` and `time_left` where the result is almost never a round number. This is needed to preserve precision in Python 2.
When converting the result to an integer again we should use `int(round(result))`.

The current implementation suffers from rounding errors in both Python 2 and Python 3.
For instance 86.99% is converted to 86%, but it should be converted to 87% to represent reality more correctly.

> What should be done if the dl_size (the zipfilesize of the Chrome OS image extracted from recovery.conf) does not match total_length?

I don't know whether we should take this into account. I think it's unlikely to happen and when it happens, the download may still succeed without errors. So, I think that logging a warning message is sufficient.

horstle:
> We need to pass at least one `float` when calculating `percent` and `time_left` where the result is almost never a round number. This is needed to preserve precision in Python 2.

Actually not, since we do `from __future__ import division` .

> When converting the result to an integer again we should use `int(round(result))`.
> 
> The current implementation suffers from rounding errors in both Python 2 and Python 3.
> For instance 86.99% is converted to 86%, but it should be converted to 87% to represent reality more correctly.

That's why it wouldn't even make a difference atm. The fraction would be a float, but then `int()` is called, so the result would be the same as with all ints and without the import (in py2). I think being of by one percent or one second is not really dramatic, but adding `round()` is a no-brainer. I'll do that, soon.

> I don't know whether we should take this into account. I think it's unlikely to happen and when it happens, the download may still succeed without errors. So, I think that logging a warning message is sufficient.

At the end of the download we check the file size and compare it with `dl_size`. If that didn't match `total_length`, the check would fail, even i the download actually succeeded. I think we should log a warning and make dl_size equal to total_length.
Or should `dl_size` be removed from the function completely and only `total_length` be used?

mediaminister:
Okay, adding `round()` will fix precision.

> Or should dl_size be removed from the function completely and only total_length be used?

No, I think it is useful to detect possible differences between data from `recovery.conf` and the actual http headers. The download shouldn't fail on this but logging it will definitely help us this causes problems.

horstle:
Alright, can be reviewed/merged now.

mediaminister:
I noticed that **time_left** stays at zero while extracting and searching for the Widevine CDM. But, I guess this is not related to this PR. I saw this issue before.

horstle:
Yes, I wanted to look into that next. Maybe `progress_dialog()` doesn't inherit the `close()` method from `DialogProgress`?

PR #359, opened 2020-05-24
Wait before creating progress dialog

horstle:
If a progress dialog is created right after calling `close()` on another, it might happen that it isn't actually closed, so the content (especially line2 and line3) would be inherited. That's why the "Time remaining: 0:00" stayed there on ARM.

On Kodi 19 this issue doesn't exist, because line2 and line3 don't exist there, but there is still a warning in the debug log that a dialog is already running.


mediaminister:
I'm not yet sure this is the right fix. I should investigate.

horstle:
I know this isn't a good fix. Ideally `progress.close()` would only return after the progress was actually closed, but that's a change Kodi would have to make and until then, this is imo a good enough temporary fix.

mediaminister:
I investigated and found a better solution: https://github.com/emilsvennesson/script.module.inputstreamhelper/pull/370

PR #360, opened 2020-05-28
Changes to last_update

horstle:
This resolves the inconsistency between using last_update for when widevine was last installed and for when an update check was made.

This relates to #344.

### DIFF
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -227,7 +227,6 @@ class Helper:
 
         if self.install_and_finish(*result):
             from time import time
-            set_setting('last_update', time())
             set_setting('last_check', time())
             return True
 
@@ -242,6 +241,7 @@ class Helper:
             log(0, 'Removed Widevine CDM at {path}', path=widevinecdm)
             delete(widevinecdm)
             notification(localize(30037), localize(30052))  # Success! Widevine successfully removed.
+            set_setting('last_modified', '0.0')
             return True
         notification(localize(30004), localize(30053))  # Error. Widevine CDM not found.
         return False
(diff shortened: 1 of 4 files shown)


lib/inputstreamhelper/__init__.py
```diff
@@ -301,7 +301,7 @@ def _update_widevine(self):
                 log(3, 'User declined to update {component}.', component=component)
         else:
             from time import mktime
-            set_setting('last_update', mktime(datetime.utcnow().timetuple()))
+            set_setting('last_update_check', mktime(datetime.utcnow().timetuple()))
```
dagwieers, line 304: Can we not call this simply `last_check` ? It is less confusing IMO.

resources/settings.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <settings>
     <setting id="last_update" value="0.0" visible="false"/>
+    <setting id="last_update_check" value="0.0" visible="false"/>
```
dagwieers, line 4: I think it is weird that these are marked as floats, but I guess the timestamp returns sub-second timestamps?

lib/inputstreamhelper/__init__.py
horstle: Agree and done.

resources/settings.xml
horstle: I didn't really think about that, but yes, mktime returns a float.

horstle: Just had another look. The timetuple doesn't have ms so it should always be .0 and it could be saved as int. But there is no real need for that.

horstle: I added a commit which (imho) simplifies the code a little. With this last_update and last_check are "real" floats.

horstle:
Alright, I think there is nothing to add here. Please have a look, whether this can be merged.

mediaminister:
I found a problem:
- When you manually install or remove Widevine and directly switch to the information window , the widevine version is immediately updated, but the date is only updated after closing en reopening InputStreamHelper settings.

horstle:
Yes, we had that issue before. It's because we are changing a setting from the settings page, which si only stored after closing it. It's a Kodi issue and there's nothing we can do about it.

mediaminister:
This can be merged.

dagwieers:
Ok, I will propose to close the settings when updating or removing Widevine. This will avoid the issue reported. If both operations show progress bars. I have some other smaller things we could improve.

PR #363, opened 2020-06-07
Changes to Chrome OS recovery

horstle:
This changes recovery.conf to recovery.json, which is easier to load (see changes to `chromeos_config()`) and should be more future proof, since recovery.json is also used by the "Chromebook recovery utility" addon for Chrome browser.
Additionally it also only saves the data of the chosen image instead of the whole recovery.json, making it easier to read (and saving a tiny bit of diskspace).

### DIFF
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -281,7 +281,7 @@ class Helper:
             current_version = wv_config['version']
         else:
             component = 'Chrome OS'
-            current_version = select_best_chromeos_image(wv_config)['version']
+            current_version = wv_config['version']
 
         latest_version = latest_widevine_version()
         if not latest_version:
@@ -436,8 +436,7 @@ class Helper:
             if arch() in ('arm', 'arm64'):  # Chrome OS version
                 wv_cfg = load_widevine_config()
                 if wv_cfg:
-                    installed_img = select_best_chromeos_image(wv_cfg)
-                    text += localize(30822, name=installed_img['hwidmatch'].split()[0].lstrip('^'), version=installed_img['version']) + '\n'
+                    text += localize(30822, name=wv_cfg['hwidmatch'].split()[0].lstrip('^'), version=wv_cfg['version']) + '\n'
             if get_setting_float('last_check', 0.0):
                 wv_check = strftime('%Y-%m-%d %H:%M', localtime(get_setting_float('last_check', 0.0)))
             else:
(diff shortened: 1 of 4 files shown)


dagwieers:
I don't mind doing this, but there is a benefit to keep all the ChromeOS images in the `recovery.json` for debugging or troubleshooting. It may help people understand better what is going on and why.

That said the `recovery.json` can be downloaded, but it is always a point in time. So to understand behaviour you need the point-in-time that reflects the log-files. Otherwise you have to ask to try again and hope it fails identically. Having the verbatim file around avoids this.

mediaminister:
> Additionally it also only saves the data of the chosen image instead of the whole recovery.json, making it easier to read (and saving a tiny bit of diskspace).

I don't think this is an improvement. Having a full backup of the recovery.json may help us when InputStreamHelper is completely broken because of a breaking change at Google.

horstle:
Yes, we as developers should have a copy of (multiple versions of) the recovery.json, but I don't see the point in keeping the whole thing on every users system. Since #351 I have a script running, which checks the recovery.json twice a day and sends it to me via email, if it changed. Maybe we could implement something similar as Github action. Keeping the whole recovery on a normal users system just makes it unnecessary complicated to find out the version and name of the image used for extraction.

mediaminister:
> but I don't see the point in keeping the whole thing on every users system.

Because we can't redistribute copyrighted files. So, we can't keep this file in our own public repo, the user should get this file directly from Google servers.

Eventually, if this file is cached at the users system, we may use this file when Google is down...

I don't oppose merging this change, but I don't think that caching less information is an improvement.

horstle:
> Because we can't redistribute copyrighted files. So, we can't keep this file in our own public repo, the user should get this file directly from Google servers.

I still can't think of any situation where it would help to keep the information about all the images. If you have any example, please explain.

> Eventually, if this file is cached at the users system, we may use this file when Google is down...

But after this change none of the information about the chosen image is lost. How could the information about all the other images possibly help?

> I don't oppose merging this change, but I don't think that caching less information is an improvement.

It definitely is an improvement, if you look at all the simplified calls to load_widevine_config().

mediaminister:
> How could the information about all the other images possibly help?

If suddenly the currently used Widevine ARM version doesn't work anymore and Google doesn't provide that certain image anymore and also changed the structure of the recovery.json.
I know, a worst case scenario, but it's possible.

horstle:
In that case, we'd have to find out which of the other images is still available (most probably they would all be gone simultaneously, but let's just assume they aren't) and has a working widevinecdm. Then we'd have to hard-code the download url until we find a different solution. Having the complete recovery.json on every user system still wouldn't help.
And this is extremely unlikely, since the official recovery tool uses the recovery.json (thus the move from conf to json).

That said, how about we save the complete recovery.json and another json file with the information of the chosen image? `load_widevine_config()` could use that different file and still be simpler, yet we'd still have the recovery.json for whenever it might help.

mediaminister:
> That said, how about we save the complete recovery.json and another json file with the information of the chosen image?

Perfect!

horstle:
Alright, I changed the second commit accordingly.

PR #367, opened 2020-06-09
Update French translation

mediaminister:


horstle:
With my limited knowledge in french, looks good to me.

PR #369, opened 2020-06-12
libwidevinecdm.so exists twice in CrOS images

### LINKED ISSUE #368: IndexError when extracting Widevine CDM from Chrome OS
I was testing the current master branch and I got the following `IndexError`:
```
2020-06-12 10:42:27.021 T:140529528923904   ERROR: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                             - NOTE: IGNORING THIS CAN LEAD TO MEMORY LEAKS!
                                            Error Type: <type 'exceptions.IndexError'>
                                            Error Contents: list index out of range
                                            Traceback (most recent call last):
                                              File "/home/.kodi/addons/script.module.inputstreamhelper/default.py", line 8, in <module>
                                                run(sys.argv)
                                              File "/home/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/api.py", line 14, in run
                                                widevine_install()
                                              File "/home/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/api.py", line 42, in widevine_install
                                                Helper('mpd', drm='widevine').install_widevine()
                                              File "/home/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/__init__.py", line 34, in clean_before_after
                                                result = func(self, *args, **kwargs)
                                              File "/home/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/__init__.py", line 224, in install_widevine
                                                result = install_widevine_arm(backup_path())
                                              File "/home/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/widevine/arm.py", line 160, in install_widevine_arm
                                                progress=progress)
                                              File "/home/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/widevine/arm_chromeos.py", line 106, in extract_file
                                                blk_group = self.blk_groups[blk_group_num]
                                            IndexError: list index out of range
                                            -->End of Python script error report<--
```

horstle:
This fixes #368.

The problem is that "libwidevinecdm.so" is somehow found twice in the Chrome OS image. I had actually noticed that before, but forgot about it again, otherwise I would have added this check earlier.
The other occurrence so far always was the second one, but now it was the first.

This change checks if the inode number is plausible and the name length matches and continues searching if it doesn't.

### DIFF
--- a/lib/inputstreamhelper/widevine/arm_chromeos.py
+++ b/lib/inputstreamhelper/widevine/arm_chromeos.py
@@ -91,16 +91,16 @@ class ChromeOSImage:
 
             chunk = chunk1 + chunk2
             if bin_filename in chunk:
-                break
+                i_index_pos = chunk.index(bin_filename) - 8
+                dir_dict = self.dir_entry(chunk[i_index_pos:i_index_pos + len(filename) + 8])
+                if dir_dict['inode'] < self.sb_dict['s_inodes_count'] and dir_dict['name_len'] == len(filename):
+                    break
             chunk1 = chunk2
             if percent8 < 240:
                 percent8 += 1
                 self.progress.update(int(percent8 / 8))
 
         self.progress.update(32, localize(30062))
-
-        i_index_pos = chunk.index(bin_filename) - 8
-        dir_dict = self.dir_entry(chunk[i_index_pos:i_index_pos + len(filename) + 8])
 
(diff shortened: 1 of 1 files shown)


mediaminister:
Thanks for fixing this. It works again.

PR #370, opened 2020-06-12
Wait for previous Progress dialog to close

mediaminister:
If a progress dialog is initialized right after calling close() on another, it might happen that the previous progress dialog is still active and text from the previous dialog is displayed in the new one.

Checking if there is a progress dialog active and waiting until it closes before initializing a new one solves this problem.

### DIFF
--- a/lib/inputstreamhelper/kodiutils.py
+++ b/lib/inputstreamhelper/kodiutils.py
@@ -16,6 +16,14 @@ ADDON = xbmcaddon.Addon('script.module.inputstreamhelper')
 class progress_dialog(DialogProgress, object):  # pylint: disable=invalid-name,useless-object-inheritance
     """Show Kodi's Progress dialog"""
 
+    def __init__(self):
+        """Initialize a new progress dialog"""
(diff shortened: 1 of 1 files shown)


lib/inputstreamhelper/kodiutils.py
```diff
@@ -16,6 +16,16 @@
 class progress_dialog(DialogProgress, object):  # pylint: disable=invalid-name,useless-object-inheritance
     """Show Kodi's Progress dialog"""
 
+    def __init__(self):
+        """Initialize a new progress dialog"""
+        # Wait for previous Progress dialog to close
+        # Progress dialog Window ID is 10101: https://kodi.wiki/view/Window_IDs
+        previous_active = get_current_window_id() == 10101
+        while previous_active:
```
horstle, line 24: All in all a good solution, just one thing about the style:
Why don't you put `while get_current_window_id() == 10101:` here and remove both lines defining `previous_active`?

mediaminister: Thanks for noticing, that's much better.

PR #371, opened 2020-06-13
Translate new added strings to Romanian

tmihai20:
Translated to Romanian new added strings


mediaminister:
Many thanks for the translation.

PR #375, opened 2020-06-14
Update german translation for version 0.5.0

tweimer:


resources/language/resource.language.de_de/strings.po
```diff
@@ -235,73 +235,73 @@ msgstr "Keine Backups gefunden!"
...
 ### INFORMATION DIALOG
 msgctxt "#30800"
 msgid "[B]Kodi[/B] version [B]{version}[/B] is running on a [B]{system}[/B] system with [B]{arch}[/B] architecture."
-msgstr ""
+msgstr "[B]Kodi[/B] Version [B]{version}[/B] läuft auf einem [B]{system}[/B] System mit [B]{arch}[/B] Architektur."
 
 msgctxt "#30810"
 msgid "[B]InputStream Helper[/B] is at version [B]{version}[/B] {state}"
-msgstr ""
+msgstr "[B]InputStream Helper[/B] ist in Version [B]{version}[/B] {state}"
```
horstle, line 280: I think "... ist __bei__ Version ..." sounds better. Same in the following lines.

tweimer: I disagree. Never heard "bei Version" anywhere.

mediaminister: In Dutch it's translated to "is versie", so maybe "ist Version" in German is a good solution?

horstle: I understand why that would make sense to you, but no, "ist Version" wouldn't be correct. One possibility would be to use "hat Version", although that technically translates to "has version" in english, not "is at version", but the meaning would still be the same.
Another alternative might be "ist auf Version", but to me, "ist in Version" only works in contexts like "A ist in Version X erschienen/installiert/verfügbar...", but not here.

dagwieers: Being involved in translations more than ever before, I find some of the correct constructs in Dutch feel very wrong. It is weird to feel alienated to my own language simply because I never use Dutch for technology :-)

horstle: Yes, that might be part of my problem, too. I searched around for a while now, but can't find any grammatical reason for/against any of the prepositions.

One could say, that "at" in english is usually translated to "bei" in german, but then again there are situation where different prepositions are used, e.g. "on" would usually be translated to "auf", but "to be __on__ the bus" in german isn't "__auf__ dem Bus sein", but "__im__ [short for "__in__ dem"] Bus sein".

As a last resort, I had a look at how [deepl.com](https://www.deepl.com/translator#en/de/A%20is%20at%20version%20B) would translate it. Although the proposed translation is "ist bei", "ist in" is also an option...

tweimer: Well, I would always say e.g. "Ich  nutze Thunderbird in Version 68.9", never "bei". "Bei" sounds weird here.

horstle: Yes, "bei" would sound wrong there, but it sounds much better than "in" to me in the given situation. I assumed this would be the case for most people, so I didn't expect such a lengthy discussion about this.

If you are completely unwilling to make any of my proposed changes (hat, ist auf, ist bei), I'll just merge it as is this evening or tomorrow morning, since it doesn't really sound wrong to me, just strange. I just think that any of the others would sound better.

tweimer: Well, after reading your suggestions again, I'm still "completly unwilling" to change it, sorry. It's a matter of opinion.

PR #273, opened 2020-02-22
Do a cleanup before install…

horstle:
, since leftovers from a previously failed install might cause issues with downloads.

See https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/265#issuecomment-589925898

PR #379, opened 2020-06-20
Optimize file operations

mediaminister:
This pull request adds optimizations to reduce the load on the file system:
- bugfix: add trailing file system separator (slash/backslash) to folder paths because `xbmcvfs.exists` needs this to succesfully check if a folder exists
- only create mount path when it will be used

### DIFF
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -185,7 +185,7 @@ class Helper:
         if downloaded:
             progress = progress_dialog()
             progress.create(heading=localize(30043), message=localize(30044))  # Extracting Widevine CDM
-            unzip(store('download_path'), os.path.join(bpath, cdm_version))
+            unzip(store('download_path'), os.path.join(bpath, cdm_version, ''))
 
             return (progress, cdm_version)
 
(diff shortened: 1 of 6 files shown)


lib/inputstreamhelper/__init__.py
```diff
@@ -26,15 +26,14 @@ class InputStreamException(Exception):
 
 
 def cleanup_decorator(func):
-    """Decorator which runs cleanup before and after a function"""
+    """Decorator which runs cleanup after a function"""
 
-    def clean_before_after(self, *args, **kwargs):  # pylint: disable=missing-docstring
+    def clean_after(self, *args, **kwargs):  # pylint: disable=missing-docstring
         # pylint only complains about a missing docstring on py2.7?
-        self.cleanup()
```
horstle, line 33: We had some rare situations previously, where ish would crash during install, which meant it didn't clean up. That wouldn't be a big issue by itself, but if the bin was partially extracted, or still mounted, when ish ran again, that could cause issues. See e.g. https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/265#issuecomment-589849360 and the resulting PR #273. Also note that an unmount before/at the beginning of the install was already there before my change.

I think none of these issues could arise with the new ARM install method, so a cleanup before might not be necessary there, but while we still have the legacy ARM install method, I think this should stay. I was planning to remove the legacy method a few releases later, when the new one tried and tested. IMO the cleanup before can be removed then, too, since the only thing in temp will be the downloaded zips (the chrome os and the x86 for the license) and I think they would simply be overwritten, if they were already there.

PR #380, opened 2020-06-20
v0.5.0 release

mediaminister:


emilsvennesson:
I'd also add @horstle to the authors list.

horstle:
Thanks 😃

mediaminister:
> Not sure if we need some more time for testing the most recent changes.

I tested on Windows 10 with Kodi 18/19, on LibreELEC VM's running ARM and X86 flows and Kodi18/19 on Linux x86 and on Android VM. It never crashed or caused problems.

I'll release tomorrow.

PR #382, opened 2020-06-20
GRstrings for 0.5.0

Twilight0:
Filled #30058, #30059, #30060, #30061, #30062, #30063, #30064, #30065, #30800, #30810, #30811, #30820, #30821, #30822, #30823, #30824


mediaminister:
Thanks for the translation.

PR #393, opened 2020-09-17
Fix typo in dutch translation

michaelarnauts:
`{version}` was wrongly translated. This causes the information dialog to not show the version.


horstle:
Thanks for the fix.

PR #400, opened 2020-10-02
Log when we download a new ChromeOS ARM image

dagwieers:
This is crucial for our ARM users to understand a new Widevine was installed from a specific ChromeOS image.

Having this will make it much easier for users to idenfity the cause of their recent problems, as well as help us identify the affected ChromeOS image in case of issues.


dagwieers:
We tend to log as little as possible already, but this event happens maybe once every 31 days so I think this is warranted.

horstle:
I'm fine with logging it, but in general I prefer ".format" over "%". So I'd do something like `"...".format(**arm_device)`.

horstle:
lgtm, I'll merge it.

PR #401, opened 2020-10-03
Update Chrome OS ARM hardware id's

mediaminister:
This PR includes:
- add new ARM hardware id's
- remove all end-of-life ARM hardware id's
- remove router hardware id's because they lack the widevine binary
- Slightly improve hardware id selection in `arm.py`

I used the following resources to update the list of hardware id's:
https://www.chromium.org/chromium-os/developer-information-for-chrome-os-devices
https://cros-updates-serving.appspot.com/

### DIFF
--- a/lib/inputstreamhelper/config.py
+++ b/lib/inputstreamhelper/config.py
@@ -79,36 +79,24 @@ WIDEVINE_CONFIG_NAME = 'manifest.json'
 
 CHROMEOS_RECOVERY_URL = 'https://dl.google.com/dl/edgedl/chromeos/recovery/recovery.json'
 
-# Last updated: 2019-08-20 (version 12239.67.0)
+# To keep the Chrome OS ARM hardware ID list up to date, the following resources can be used:
+# https://www.chromium.org/chromium-os/developer-information-for-chrome-os-devices
+# https://cros-updates-serving.appspot.com/
+# Last updated: 2020-10-05
 CHROMEOS_RECOVERY_ARM_HWIDS = [
-    # 'ARKHAM',
-    'BIG',
-    'BLAZE',
     'BOB',
-    # 'DAISY',
     'DRUWL',
     'DUMO',
     'ELM',
-    'EXPRESSO',
     'FIEVEL',
     'HANA',
-    'JAQ',
-    'JERRY',
+    'JUNIPER-HVPU',
     'KEVIN',
-    'KITTY',
+    'KODAMA',
+    'KRANE-ZDKS',
     'MICKEY',
-    'MIGHTY',
-    'MINNIE',
-    'PI',
-    'PIT',
-    'RELM',
     'SCARLET',
-    'SKATE',
-    'SNOW',
-    'SPEEDY',
-    'SPRING',
     'TIGER',
-    # 'WHIRLWIND',
 ]
 
 CHROMEOS_BLOCK_SIZE = 512


dagwieers:
Well done. You beat me to it, and those reference will help us in the future as well!

Could we add these to the code as comments? It provides transparency for others, and may get us contributions in the future from outsiders.

(Originally, I looked up every model to check the reported architecture on technology sites. This is why the errors were introduced IMO 😕)

dagwieers (approved):

dagwieers:
With this new list we may want to release v0.5.2 too.

@emilsvennesson Can we automate our release workflow in this project, please?

emilsvennesson:
Great job getting this fixed so quickly!

@dagwieers Yeah, of course! Could I get some pointers what's needed from my side to get it up and running?

dagwieers:
@emilsvennesson We need 3 secrets in this GitHub repository:
- EMAIL
- GH_USERNAME
- GH_TOKEN 

See https://github.com/xbmc/action-kodi-addon-submitter for more information.

horstle (changes requested):
I had a thorough look at the linked sites and simultaneously at the current recovery.json and found a few more issues:

* RELM is x86, too, so should be removed
* The hwidmatch in recovery.json for JUNIPER is JUNIPER-HVPU, KRANE is KRANE-ZKDS
* Also, JUNIPER comes after JERRY in the alphabet ;)
* Currently select_best_chromeos_image expects a whitespace after the name in the recovery.json (e.g. `"hwidmatch": "^BOB .*"`), but that's not the case for KODAMA (`"hwidmatch": "^KODAMA.*"`), which doesn't really matter, since it's the same image as KRANE/KRANE-ZKDS anyway.
* Similarly DRUWL, DUMO and SCARLET use the same image, but I'd still keep them all, in case one might not get updated or get a different image in the future (although I don't really expect that to happen). Maybe add a comment that they are (currently) the same.

mediaminister:
> Well done. You beat me to it, and those reference will help us in the future as well!
> 
> Could we add these to the code as comments? It provides transparency for others, and may get us contributions in the future from outsiders.

Thanks, I added the resources as a comment.

> I had a thorough look at the linked sites and simultaneously at the current recovery.json and found a few more issues

Thanks, I fixed the issues.

dagwieers:
I  went over the lists, and found that:
- BIRCH, HANAWL, JUNIPER, KRANE MAPLE, MAPLE14, SYCAMORE and SYCAMORE360 are missing (also from recovery.conf)
- I could not find SCARLET (it is a board name, not a codename)
- JAQ, JERRY, KITTY, MICKEY, MIGHTY, MINNIE and SPEEDY are EOL as well (everything up to 2015, ELM is 2016 and is ok)
- Although MICKEY (2015) does still have the latest version delivered

lib/inputstreamhelper/widevine/arm.py
```diff
@@ -79,7 +79,7 @@ def select_best_chromeos_image(devices):
     for device in devices:
         # Select ARM hardware only
         for arm_hwid in config.CHROMEOS_RECOVERY_ARM_HWIDS:
-            if '^{0} '.format(arm_hwid) in device['hwidmatch']:
+            if device['hwidmatch'].startswith('^{0}'.format(arm_hwid)):
```
dagwieers, line 82: This is going to be problematic. The reason why we added the whitespace is because if you match for `^BOB`, you will also get `^BOBBA` and `^BOBBA360`, which are both `x86_64`.

dagwieers (changes requested):

lib/inputstreamhelper/widevine/arm.py
mediaminister: Thanks, fixed it.

horstle:
Yes, I saw BIRCH, HANAWL, MAPLE, MAPLE14, SYCAMORE, SYCAMORE360 and TELESU are missing, too, but didn't mention it, because as you said they are not in recovery.json and also, they all have board name "hana", so I guess it would be the same image as HANA anyway.
Even if they had different images, adding them to the list might not make sense yet, because they could well be called differently, like JUNIPER is JUNIPER-HVPU or KRANE is KRANE-ZDKS (btw great job finding/fixing my typo there @mediaminister).
SCARLET also uses the same image as DRUWL and DUMO, so again doesn't really matter if it's there or not.

I think it might actually make more sense to look at the name in "file" (after the second underscore) since they are afaics completely consistent with the board names in the chromium.org list (except for "_" in the list being "-" in the json). Then we wouldn't need to discuss whether to put all the "hana"s and all the "scarlet"s in the list or not.

emilsvennesson:
Just a thought: maybe we should move the ARM HWID:s to its own JSON file and load them from there? We could then compare with the master and refresh them if they're out of sync. It would allow us to update the list without having to push out a new update.

dagwieers:
> Just a thought: maybe we should move the ARM HWID:s to its own JSON file and load them from there? We could then compare with the master and refresh them if they're out of sync. It would allow us to update the list without having to push out a new update.

I am not a big fan of hosting this file because it is one more operation that may break unexpectedly _and_ I expect this mechanism to disappear over time. This is the first time we modified the HWID list in 13 months, and it was because it contained errors, not because the list required to be updated.

What I would suggest we do is update the list every time we intend to do a new release anyway. And we could also decide to do a release every 6 or 12 months to update the list if no other release was made. (Note that until new maintainership the longest time between releases was 5 months).

dagwieers:
Let me add that Chrome v87 is slated to have ChromeOS and Chrome split up (LaCros). Which most likely means Widevine will no longer ship with the ChromeOS image anymore and we will have to implement another means. Most likely we will have to fix our implementation to use an older image (by version) until we have an alternative in place.

And as it is now, we cannot check for v87 from recovery.conf

mediaminister:
I think it only makes sense to list ARM hardware id's in `config.py` that are defined in `recovery.conf`. So TELESU is not listed and SCARLET is listed.

I removed the additional EOL ARM hardware id's.

I agree that there is no need to complicate this logic. The reason it broke was the fact there were several x86_64 hardware id's in the list.
The list will continue to work if we make sure we list at least one ARM hardware id that is available in recovery.json
Currently, we have 13 ARM hardware id's in config.py, so I'm confident this will work as long as Google continues to support the recovery.json

dagwieers (approved):
Looks good to me.

dagwieers:
> The list will continue to work if we make sure we list at least one ARM hardware id that is available in `recovery.json`

A popular alternative script relies on `ELM` only and it needs to uncompress the image. We do much better on both accounts ;-)

horstle (approved):

dagwieers:
Not sure if we want to plan a v0.5.2 release for this. Do we have anything else to go with it in the pipeline?
Maybe call out for translations once more?

We do have a few Enhancements/Bugs open. Should we discuss a new roadmap with ideas for the future?
https://github.com/emilsvennesson/script.module.inputstreamhelper/wiki/Ideas

mediaminister:
I think we should investigate and estimate (if possible) what the impact will be of the following two future events:
- The release of ARM-based Apple Macintosh hardware by the end of October (and the release of a MacOS ARM64 Widevine CDM)
- The release of Chrome OS 87 by the end of November (https://chromiumdash.appspot.com/schedule)

A release of v0.5.2 or higher can be done sometime in November.

dagwieers:
@horstle I would like to invite you to an internal Slack, but I would need an email-address to do this. Can you send me one in private? (See Github account for contact details)

horstle:
I just send you an email.

dagwieers:
After testing ChromeOS v87 canary images it is clear that _libwidevinecdm.so_ is still included in ChromeOS v87.

PR #405, opened 2020-10-18
[python] update xbmc.translatePath() to xbmcvfs

heitbaum:
[python] update xbmc.translatePath() to xbmcvfs


dagwieers:
Thanks, I modified this slightly and made the tests work.

Issue #388, opened 2020-07-05
Invalid setting type


horstle:
I recently saw a debug log of a standard LibreELEC installation and it had a line:
`EXCEPTION: Invalid setting type`

It seems that the call to `getSettingNumber` in `get_setting_float` fails with a type error, when called on `last_check` or `last_modified`, e.g. in `info_dialog`. That's not really a problem, since it is handled by `get_setting_float`, but still should be looked into.
Even after changing `getSettingNumber` to `getSettingInt` and changing both values to a small integer it still produced the same error, but when I changed the lines in `info_dialog` to read the values of `update_frequency`, the error vanished, even though it looked exactly the same in the settings.xml.

When I realised this I also thought again about the problem that both these values aren't stored if installing from the settings page and exiting the page with the "Abort" button afterwards. So I thought maybe we should create an independent xml/json/... file to store those values instead of hiding them in the settings. That would avoid both these issues. What do you think?

dagwieers:
I would prefer to keep it in the *settings.xml* if possible. I guess we have to figure out what is going on and if needed, get this fixed upstream if it is a bug...

horstle:
Ok, some more things I tried:
* Set the values with setSettingNumber - that didn't change anything
* Specify `type="number"` in the settings.xml - after that they couldn't be read at all and the settings page had an additional tab...

mediaminister:
I found out two things:
- `type="number" `can only be used for integers, so `getSettingNumber` is useless.
- invisible settings  with` type="number"` need to be in a category otherwise an empty tab is shown.

PR #410, opened 2020-11-10
Fix invalid setting type

mediaminister:
This fixes https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/388

I guess `xbmcaddon.getSettingNumber` isn't functional because you can only define an integer type in the Kodi addon settings xml and in addition it is wrongly named as `type="number"`.

Maybe this can be fixed upstream if Team Kodi still accepts changes to the deprecated old settings format:
https://kodi.wiki/view/Add-on_settings#type.3D.22number.22

Relevant source code
https://github.com/xbmc/xbmc/blob/5230b683323ca58c62459a371c1306a6cb4d4644/xbmc/addons/settings/AddonSettings.cpp#L832-L844

`setting->GetType() `can never be a `SettingType::Number`, so it differs from `TSetting::Type()` and returns false
https://github.com/xbmc/xbmc/blob/c4de26b28fb13ce65eb1caac4236d83bd410d112/xbmc/addons/Addon.cpp#L206

### DIFF
--- a/lib/inputstreamhelper/kodiutils.py
+++ b/lib/inputstreamhelper/kodiutils.py
@@ -193,14 +193,11 @@ def get_setting_int(key, default=None):
 
 def get_setting_float(key, default=None):
     """Get an add-on setting as float"""
+    value = get_setting(key, default)
     try:
-        return ADDON.getSettingNumber(key)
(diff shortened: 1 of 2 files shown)


horstle (approved):
LGTM. I didn't test it, but I can't see how it could break anything and the tests seem to be successful.

Issue #248, opened 2020-01-15
Suggestion/request to information screen


CastagnaIT:
Many, many, many people do not understand that on android you can not get the Widevine version with ISHelper
They read "Not found" and most of the time always think the component _is missing_.

I'm really tired of saying the same thing to every new person who comes in...

![image](https://user-images.githubusercontent.com/3257156/72441318-c4909880-37aa-11ea-995d-97e07a15828d.png)

I suggest adding a variable to identify the execution on Android system
and add a new line of information such as:
`- Widevine package:` Embedded [on android] / Installed [to others] / Not installed

And for this line:
`- Widevine version:` N° of version
make it non-visible on android system

this would make it easier for everyone to understand and read the info data

dagwieers:
This is already fixed in the master branch, but we haven't released it yet.

CastagnaIT:
oh luckily, well
thanks!

dagwieers:
Given that we did this in October 2019, maybe we ought to consider making a new v0.5.0 release?

dagwieers:
I would also like to query the library to get the version from it directly, rather than grep for a pattern.

CastagnaIT:
you can try to get the android widevine version in in this way:

create a instance of xbmcdrm.CryptoSession, then get the "version" string

look at android_crypto.py in https://github.com/CastagnaIT/plugin.video.netflix/pull/438

dagwieers:
@CastagnaIT It doesn't seem to work, every key I ask for is empty. At least on a Raspberry Pi.

dagwieers:
This works to get the version on Linux:
```python
from ctypes import *
_widevine_lib = cdll.LoadLibrary('libwidevinecdm.so')
get_cdm_version = _widevine_lib.GetCdmVersion
get_cdm_version.restype = c_char_p
print('Widevine version: %s' % get_cdm_version())
```

PR #415, opened 2020-12-23
Improve get_lib_version, use as check

### LINKED ISSUE #413: Request to better manage case of failed widevine update for Linux ARM
Several times it has happened that users go crazy when a widevine update is automatically performed with a different architecture than ARM, and i know that the wv library search cannot always be perfect

this rightly breaks all addons that need this library, but the problem is that users never know what to do
and they open multiple Issue requests everywhere

i have noticed that the "ldd" command in inputstreamhelper always fails when this problem happen

My suggestion are:
- Use the "ldd" command to check if the downloaded library is compatible with the system, this before perform the update and break all add-ons. So if the command return error avoid the update and inform the user on what to do
- A better message for this case of problem, with reference to what to do (like read GitHub Wiki FAQ)

I hope something will be done about it because it's so annoying for both the add-ons maintainers and the users who don't know what to do

horstle:
Based on the code from @CastagnaIT, this improves the functionality of get_lib_version and adds a check to it. It now returns an empty string if the file is corrupt or the architecture doesn't match. 

Currently unit tests fail. I guess that's because they are running on x86 hardware, so the aforementioned check fails. So far I only tested Linux x86 manually.

Closes #413.

### DIFF
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -11,7 +11,7 @@ from .kodiutils import (addon_version, delete, exists, get_proxies, get_setting,
                         set_setting, set_setting_bool, textviewer, translate_path, yesno_dialog)
 from .utils import arch, http_download, remove_tree, run_cmd, store, system_os, temp_path, unzip
 from .widevine.arm import install_widevine_arm, unmount
-from .widevine.widevine import (backup_path, has_widevinecdm, ia_cdm_path, install_cdm_from_backup, latest_widevine_version,
+from .widevine.widevine import (backup_path, has_widevinecdm, get_lib_version, ia_cdm_path, install_cdm_from_backup, latest_widevine_version,
                                 load_widevine_config, missing_widevine_libs, widevine_config_path, widevine_eula, widevinecdm_path)
 from .unicodes import compat_path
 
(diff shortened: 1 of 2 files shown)


mediaminister:
> Currently unit tests fail.

Strange, maybe related to this: https://github.community/t/segmentation-fault-when-running-binary-in-a-job/18418

mediaminister:
@horstle 
I found and fixed the problem:
- the segmentation fault was caused by loading the shared library several times in quick succession. Unloading the library fixes this.
There was also a problem with loading the 32-bit ARM lib. It throws an `OSError` with `wrong ELF class: ELFCLASS32`

I added the old `get_lib_version` again as a fallback.

Feel free to make a better fix, but this is good enough for me to get merged.

dagwieers:
Ha, while looking at previous discussions I found my original code from a year ago:

https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/248#issuecomment-575961788

horstle:
@mediaminister It is __supposed__ to fail when there is an arch mismatch, so it won't install the wrong lib. That's what this PR is trying to fix (see #413). Also, the legacy version doesn't detect an arch mismatch, so it can't be used as a check.
Instead of writing worse code to pass the tests, we should improve the tests.

I'm currently looking into AWS CodeBuild, so we can run the ARM tests on ARM hardware. I think it makes sense to have separate Github actions for each platform, then the ARM tests could be run via CodeBuild and maybe at some point the Windows tests could run on Windows (it seems that's already possible in Github, but idk if it would cost something, might look into it later).

mediaminister:
Okay, I removed the fallback. Tests fail, but now you can see why.

Issue #414, opened 2020-12-15
Track ChromeOS image selection and periodically update the HWID list


dagwieers:
This is vaguely related to #413 

We probably want to periodically make a release with an updated HWID list.

This could have two advantages:
- We release fixes/improvements on a regular schedule (e.g. every 3 or 4 months)
- We stay on top of changes (e.g. does ChromeOS v89 still ship a Widevine CDM?)

Also, it would be nice if we can keep track of which ChromeOS image is selected at any given time so that if there is a change in the image (e.g. MICKEY is no longer updated) we know this may impact users and an update is forced in the next 31 days (default).

PR #416, opened 2021-01-03
Check Chrome OS recovery images availability

mediaminister:
This PR ads a daily status check to the github workflow to determine Chrome OS recovery image availability and includes the following features:
- status check fails when inputstreamhelper config contains missing or outdated(EOL) hardware id's
- status check succeeds when  inputstreamhelper config is up to date
- shows a message with the hardware id with  the smallest recovery image

This should fix https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/414

### DIFF
--- a/.github/workflows/status.yml
+++ b/.github/workflows/status.yml
@@ -1,0 +1,29 @@
+# This workflow checks Chrome OS recovery images availability status
+
+name: Status
+on:
+  pull_request:
+  push:
+  schedule:
+  - cron:  '0 5 * * 3'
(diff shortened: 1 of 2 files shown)


horstle:
In general it looks good, but I don't think it has to be done so often. Maybe once a week would be enough.

mediaminister:
Okay, I changed it to check every Wednesday morning.

dagwieers:
Somehow I missed this PR, but looking good !

PR #417, opened 2021-01-03
Update Chrome OS ARM hardware id's

mediaminister:


horstle:
Looks fine. KAPPA is not in any of the two resources we use, but since your script in #416 searches them for the corresponding board (and image) which is the only thing that matters, I think that's fine.

mediaminister:
KAPPA is the hwid of the recent HP Chromebook 11a: https://chromeunboxed.com/hp-debuts-companys-first-mediatek-powered-chromebook-for-259/

PR #438, opened 2021-05-10
Temporary fix for Widevine on ARM hardware

mediaminister:


dagwieers:
Well, we need to think about 2021-05-31 anyway. The best course of action is to hardcode to not download a new Widevine (until that day), and force a new download as soon as we know this is enforced.

So 2 new releases. One that disables ISH without disabling it, and one release that we ask to keep in the pipeline until we know it is enforced.

Issue #69, opened 2019-08-16
Detect that the used Widevine CDM is revoked


dagwieers:
If we can detect that the used Widevine CDM is revoked, we could use this as a trigger to check for an update. (If it is not working, we can't break it by trying to update the CDM).

If we would have such a mechanism, there is no longer a need to force checking for an update on upgrade. Because potentially an update of the Widevine CDM could turn a working system in a non-working system (this has happened before due to a broken released version).

dagwieers:
@emilsvennesson @mediaminister Is this at all possible?

emilsvennesson:
I'm not sure if this data is available anywhere. 

However, https://github.com/asciidisco/plugin.video.netflix/issues/465 could have been avoided, at least on x86. We can start comparing the supported Widevine API version with what's supported in inputstream.adaptive. There's a hidden setting in inputstream.adaptive, `WIDEVINE_API` we can retrieve (although I'm not sure it's been updated, ping @peak3d) and compare with `x-cdm-interface-versions` that's in the `manifest.json` file on x86.

dagwieers:
@emilsvennesson This PR is to track if we can implement an interface to detect this. My original idea was a service that monitors for these failures, and **inputstream.adaptive** would then send out specific messages for use to be used and act on.

peak3d:
WIDEVINE_API is updated, and currently version 10 is supported.
IIRC there are manifest files shipped with widevine for some platforms where the level can be picked from. For other platforms (COS IIRC) it is not the case

dagwieers:
@peak3d How could **inputstream.adaptive** get triggered if Widevine failed and an action is required?
I can foresee multiple causes:
- The Widevine CDM is no longer supported by **inputstream.adaptive** (e.g. Kodi 17)
- The Widevine CDM used is revoked by Widevine
- The Widevine CDM is missing or broken (corrupted, or wrong architecture)

Only in the two last cases we would be able to act (offer to upgrade, or reinstall).

dagwieers:
I wonder where this is coming from:
```log
2019-08-23 00:32:34.232 T:1169351536   ERROR: AddOnLog: InputStream Adaptive: Decrypt Sample returns failure!
```

peak3d:
Imo too deep level for checking, danger is big that you revoke by failure

dagwieers:
[...]

PR #441, opened 2021-05-25
Install older Widevine CDM when OS doesn't have TCMalloc support

mediaminister:
With the release of Widevine CDM 4.10.2252.0, Google uses a newer dynamic library that needs an operating system with TCMalloc support and a patched glibc to work. Support for this already implemented in LibreELEC and CoreELEC:
https://github.com/LibreELEC/LibreELEC.tv/pull/5376
https://github.com/CoreELEC/CoreELEC/pull/287

LibreELEC will release a new version supporting the new Widevine CDM: https://github.com/LibreELEC/LibreELEC.tv/pull/5399

 This PR tries to handle some problems InputStreamHelper users will run into.

This PR includes:
- propose the user to install an older Widevine CDM when `libtcmalloc_minimal.so` is not found.
- warn the user that Google will remove support for older Widevine CDM's on 31 May, 2021 when installing older CDM's

### DIFF
--- a/lib/inputstreamhelper/config.py
+++ b/lib/inputstreamhelper/config.py
@@ -67,6 +67,14 @@ WIDEVINE_MINIMUM_KODI_VERSION = {
     'Darwin': '18.0'
 }
 
+HARDCODED_CHROMEOS_IMAGE = {
+    'hwid': 'FIEVEL',
+    'url': 'https://dl.google.com/dl/edgedl/chromeos/recovery/chromeos_13505.73.0_veyron-fievel_recovery_stable-channel_fievel-mp.bin.zip',
+    'sha1': '9904e3141a2537f28469c7653c62200c1b03b057',
+    'version': '13505.73.0',
+    'zipfilesize': '1032442523'
+}
+
 WIDEVINE_VERSIONS_URL = 'https://dl.google.com/widevine-cdm/versions.txt'
 
 WIDEVINE_DOWNLOAD_URL = 'https://dl.google.com/widevine-cdm/{version}-{os}-{arch}.zip'
--- a/lib/inputstreamhelper/kodiutils.py
+++ b/lib/inputstreamhelper/kodiutils.py
@@ -65,6 +65,16 @@ def kodi_version_major():
     return int(kodi_version().split('.')[0])
 
 
+def kodi_os():
+    """Returns Kodi OS name as string"""
+    # It takes a while to get this info
+    count = 0
+    while ' (kernel: ' not in xbmc.getInfoLabel('System.OSVersionInfo') and count < 10:
+        count += 1
+        xbmc.sleep(100)
+    return xbmc.getInfoLabel('System.OSVersionInfo').split(' (kernel: ')[0]
+
+
 def translate_path(path):
     """Translate special xbmc paths"""
     return to_unicode(translatePath(from_unicode(path)))
--- a/lib/inputstreamhelper/widevine/arm.py
+++ b/lib/inputstreamhelper/widevine/arm.py
@@ -8,8 +8,8 @@ import json
 from time import time
 
 from .. import config
-from ..kodiutils import browsesingle, copy, exists, localize, log, mkdir, ok_dialog, open_file, progress_dialog, yesno_dialog
-from ..utils import cmd_exists, diskspace, http_download, http_get, run_cmd, sizeof_fmt, store, system_os, temp_path, update_temp_path
+from ..kodiutils import browsesingle, copy, exists, kodi_os, localize, log, mkdir, ok_dialog, open_file, progress_dialog, yesno_dialog
+from ..utils import cmd_exists, diskspace, http_download, http_get, http_head, run_cmd, sizeof_fmt, store, system_os, temp_path, update_temp_path
 from ..unicodes import compat_path, to_unicode
 from .arm_chromeos import ChromeOSImage
 
@@ -122,14 +122,42 @@ def chromeos_config():
     return json.loads(http_get(config.CHROMEOS_RECOVERY_URL))
 
 
+def hardcoded_chromeos_image():
+    """Gets a hardcoded ChromeOS image"""
+    arm_device = config.HARDCODED_CHROMEOS_IMAGE
+    http_status = http_head(arm_device['url'])
+    if http_status == 200:
+        return arm_device
+    return None
+
+
 def install_widevine_arm(backup_path):
     """Installs Widevine CDM on ARM-based architectures."""
-    devices = chromeos_config()
-    arm_device = select_best_chromeos_image(devices)
+    # With the release of Widevine CDM 4.10.2252.0, Google uses a newer dynamic library that needs TCMalloc support and a patched glibc to work
+    # Google will remove support for older Widevine CDM's on May 31, 2021
+    # More info at https://github.com/xbmc/inputstream.adaptive/issues/678 and https://www.widevine.com/news
+
+    # Experimental: Check for TCMalloc support
+    tcmalloc_path = '/usr/lib/libtcmalloc_minimal.so'
+    arm_device = None
+    if not exists(tcmalloc_path):
+        # Propose user to install older version
+        if yesno_dialog(localize(30066), localize(30067, os=kodi_os())):  # Your operating system probably doesn't support the newest Widevine CDM. Try older one?
+            # Install hardcoded ChromeOS image
+            ok_dialog(localize(30066), localize(30068))  # Please note that Google will remove support for older Widevine CDM's on May 31, 2021
+            arm_device = hardcoded_chromeos_image()
+            devices = arm_device
+
+    # Select newest and smallest ChromeOS image
+    if arm_device is None:
+        devices = chromeos_config()
+        arm_device = select_best_chromeos_image(devices)
+
     if arm_device is None:
         log(4, 'We could not find an ARM device in the Chrome OS recovery.json')
         ok_dialog(localize(30004), localize(30005))
         return False
+
     # Estimated required disk space: takes into account an extra 20 MiB buffer
     required_diskspace = 20971520 + int(arm_device['zipfilesize'])
     if yesno_dialog(localize(30001),  # Due to distributing issues, this takes a long time
@@ -147,7 +175,7 @@ def install_widevine_arm(backup_path):
                       localize(30018, diskspace=sizeof_fmt(required_diskspace)))
             return False
 
-        log(2, 'Downloading best ChromeOS image for Widevine: {hwid} ({version})'.format(**arm_device))
+        log(2, 'Downloading ChromeOS image for Widevine: {hwid} ({version})'.format(**arm_device))
         url = arm_device['url']
         downloaded = http_download(url, message=localize(30022), checksum=arm_device['sha1'], hash_alg='sha1',
                                    dl_size=int(arm_device['zipfilesize']))  # Downloading the recovery image
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -269,6 +269,19 @@ msgctxt "#30065"
 msgid "Shall we try again?"
 msgstr ""
 
+msgctxt "#30066"
+msgid "Warning"
+msgstr ""
+
+msgctxt "#30067"
+msgid "{os} probably doesn't support the newest Widevine CDM. Would you try to install an older Widevine CDM instead?"
+msgstr ""
+
+msgctxt "#30068"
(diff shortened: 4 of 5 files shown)


horstle:
Looks good, although I'd rather have used the time module, similar to how it's done in `update_widevine`. timestamp for June 1st would be 1622505600 seconds.

dagwieers:
So I think it is problematic that we have the date hardcoded, as we don't know whether (and when exactly) Google will be revoking the old Widevine.

So I would rather release a new package as soon as this has been detected. (And coordinate this with the repo maintainers, @basrieter @enen92 to do this without delay).

Since SCARLET was only updated now, and potentially other devices lag behind, I expect delays. Google can see what the Widevine versions are still out there, and choose a more opportune moment to make the switch.

basrieter:
If this is a one time thing for coordinating, no problem for me. We just need to discuss how you update us. I get so much mail from GitHub that just a notification will not work.

mediaminister:
> So I think it is problematic that we have the date hardcoded

Indeed, I'll change this PR to check for libtcmalloc_minimal.so and then ask the user to install the older hardcoded CDM or the newer Widevine CDM

enen92:
Ok to me. I agree with Bas, my gh email isn't even my personal email so I tend to miss stuff. Please use python-external or just DM us on slack

CastagnaIT:
i think the better thing is check the both places, the file and the `kodi.conf`, by check in some way the line:
`LD_PRELOAD=/usr/lib/libtcmalloc_minimal.so`

because an user in case of problems can disable libtcmalloc_minimal by remote line above without rollback

mediaminister:
Thanks for your suggestion. I'll change the check accordingly.

mediaminister:
@CastagnaIT Can you test this?
[script.module.inputstreamhelper-0.5.3-arm-07dad28.zip](https://github.com/emilsvennesson/script.module.inputstreamhelper/files/6552755/script.module.inputstreamhelper-0.5.3-arm-07dad28.zip)
[script.module.inputstreamhelper-0.5.3+matrix.1-arm-07dad28.zip](https://github.com/emilsvennesson/script.module.inputstreamhelper/files/6552756/script.module.inputstreamhelper-0.5.3%2Bmatrix.1-arm-07dad28.zip)

CastagnaIT:
i have see that `kodiconf_path `is hardcoded to libreelec
this check is intended only for Libreelec?

i could try to test on CoreElec but do not think to have time today

mediaminister:
I have no idea how to check for other operating systems. I guess I better remove this `kodi.conf ` check for the upcoming release. Time is limited and I should get this release out before the weekend.

Users get a yes/no dialog after the check and can decide to install the newer or older Widevine CDM.

dagwieers:
Well done. This is working as expected on my LibreELEC v18.7.

Portisch:
TCMalloc is obsolete and because of random Kodi crashes it got removed.
The problem was fixed properly in glibc dlopen itself.
No user reports so far that it does not work.

So the check for the old/new Widevine lib is not needed anymore.

https://github.com/xbmc/inputstream.adaptive/issues/678#issuecomment-855282129

horstle:
Yes, thanks. We are currently discussing what to do now, since we can't really detect if the user is on a new system with the relevant patches or not.

Portisch:
I think it's ok to remove it and "come back" to normal usage. LibreELEC did merge and remove tcmalloc also as well already.
If a user do have a old image and want to use the new CDM lib he will get a error message in kodi.log.

The correct place to "fix" this will be inputstream.adaptive to evaluate the given `dlopen` error and inform the user he need to update his system or to downgrade CDM lib. If tcmalloc is used or if glibc is patched the error will not happen.

wagnerch:
@horstle 

You could try to dlopen the library in the Python module.  Unfortunately they would have to waste the time & bandwidth to download a potentially incompatible version.

```
#!/usr/bin/env python3
import ctypes

cdm = ctypes.cdll.LoadLibrary("./libwidevinecdm.so")
GetCdmVersion = cdm.GetCdmVersion
GetCdmVersion.restype = ctypes.c_char_p
print(GetCdmVersion().decode('utf-8'))
```

Portisch:
For this you need to download the new CDM lib and the LoadLibrary will fail if glibc is not compatible.
Only way is to supply a "dummy" library with a TLS alignment of 64 bytes and test `dlopen` before downloading anything.
By this result the helper can decide if new or old CDM lib is compatible or not.

The other way is to use the check what is already implemented here:
https://github.com/xbmc/inputstream.adaptive/blob/Matrix/wvdecrypter/cdm/base/native_library_posix.cc#L31-L32

When the TLS block error happen at this place the user needs to be informed on screen to downgrade CDM lib or to update the system image to come compatible again.

wagnerch:
On thing to mention is the issue with newer widevine isn't just TLS alignment, it is also the new section with relative relocations that is unsupported by stock glibc's.  I suppose someone could craft a library like this, but the thought process for doing a "dlopen" in Python would at least give InputStream Helper the opportunity to do some recovery logic and download the older CDM, and perhaps save a setting indicating platform capabilities (knowing that this is probably a bad solution when the platform is patched, unless you intend to periodically re-test platform capabilities via 1GB download).

I think probably everyone involved would like to not have a whole bunch of tickets with CDM issues.  Pretty sure people will put tickets in for this issue against Kodi, ISA, Netflix, Amazon, and so on, it will just never end.  :)

IS Helper is probably the only hope to mitigate the numerous tickets for things that can't be solved by those maintainers.

dagwieers:
@Portisch We would like to have the ability to make inputstreamhelper reactive rather than proactive. It would allow us to only attempt updating Widevine CDM when it fails to work (e.g. when revoked) rather than after X days. See #69

Portisch:
Sure! As glibc is built from source maybe something can be added to show it's compatible, like something in the versioning or a export of a dummy function or something like this. This should work on LE/CE ARM platform.

Portisch:
Just back to the idea about glibc version:
As LE and CE do build glibc from source a string can be added to the glibc version to show if 64 byte alignment is supported or not:
https://sourceware.org/git/?p=glibc.git;a=blob;f=version.h;hb=HEAD

Then check the version like this: `ldd --version`

dtechsrv:
Okay, I see the same topic is under a closed PR.
[Unable to detect non-tcmalloc implementation of widevine fix >= 4.10.2252.0 on ARM (LE/CE) #449](https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/449)

vpeter4:
One shell script or even one indicator file in /etc could also work. Plain and simple.

mediaminister:
> Then check the version like this: ldd --version

I guess ISH should look for version 2.31.1 or higher?

mediaminister:
Please check https://github.com/emilsvennesson/script.module.inputstreamhelper/pull/450

Test packages:
For Kodi 19 Matrix: [script.module.inputstreamhelper-0.5.5+matrix.1-tls-04a0795.zip](https://github.com/emilsvennesson/script.module.inputstreamhelper/files/6662099/script.module.inputstreamhelper-0.5.5%2Bmatrix.1-tls-04a0795.zip)
For Kodi 18 Leia: [script.module.inputstreamhelper-0.5.5-tls-04a0795.zip](https://github.com/emilsvennesson/script.module.inputstreamhelper/files/6662100/script.module.inputstreamhelper-0.5.5-tls-04a0795.zip)

vpeter4:
Why would only version matter? LE/CE is using glibc-2.32 WITH extra 2 patches to allow 64 bit alignment.
I think glibc version should have some suffix to indicate if 64 bit alignment is supported or not. Like
```
# ldd --version
ldd (GNU libc) 2.32-64bittls
```
glibc 2.32 itself without patch is not useful.

Portisch:
Yes something like this. By side, it's 64 bytes, not bits.
So a version like "2.32-arm64tls" or something like this should do the job. Then the plugin can check if ARCH is arm and if yes check glibc version if compatible with new 64 byte alignment.

mediaminister:
> Why would only version matter? 

There is only a version to detect in the current LibreELEC and CoreELEC versions. I suffix could help, but this means every single os has to add a suffix.

> Only way is to supply a "dummy" library with a TLS alignment of 64 bytes
This seems the beste solution.
But I currently don't have expert knowledge to build and test such a dummy library. I don't even own ARM hardware myself. It's time consuming for me to achieve this solution.
It would be great if someone can give me a hint or documentation to build this lib.

Portisch:
I still think patching the version string is still the easiest. There are already two patches for support the new CDM lib. So just add another one patching the version. Or include it in the 64 byte alignment patch.

Then just parse the version string and the addon know if compatible or not. The patches will never be upstreamed so if anyone else want to add the support he can take this additional patch to.

I can redo the patch tomorrow to include the version string modification.

CeEs-atd:
I wonder if you could implement both ways into IS helper: 
What if IS helper checks for the "arm64tls" version of glibc and if not successful tries to dlopen an "arm64tlstest.so" library? If one of both results in success IS adaptive will able to load CDM libs with 64 byte TLS on this system. 
As soon as someone is able to create an arm64tlstest.so library that is shipped with IS helper, there would be no need to apply the version string patch to glibc any more...
The version patch is a fast fix for OS which follow fast with the glibc patch, the arm64tlstest.so a slower but more sustainable way.

dagwieers:
I don't think we want to ship a binary in a noarch add-on.

IMO inputstream.adaptive should return proper onPlayBackError information that it can act on.

It would make a lot of things easier, like only downloading a newer Widevine if the current one is revoked or fails to get loaded. (So we don't update every X days and possibly break a working setup)

Portisch:
> I don't think we want to ship a binary in a noarch add-on.

I think the same. This will become a issue.
Here the modified patch and the result how it will look like if applied:
```text
CoreELEC:~ # ldd --version
ldd (GNU libc) 2.32-arm64tls
$Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
$Written by Roland McGrath and Ulrich Drepper.
```
```c
From 1f6f179986e941f8062019fe894996550bcb737a Mon Sep 17 00:00:00 2001
From: Portisch <hugo.portisch@yahoo.de>
Date: Sat, 5 Jun 2021 19:41:25 +0200
Subject: [PATCH] tls: libwidevinecdm.so: since 4.10.2252.0 has TLS with
 64-byte alignment Change the max_align to 64U instead 16 to make it possible
 to use dlopen again. Tests by changing TLS_TCB_ALIGN directly showed up some
 random crashes. Reverence: https://lkml.org/lkml/2020/7/3/754 Modify
 'VERSION' to show TLS 64 byte alignment compatibility.

---
 elf/dl-tls.c | 5 +++++
 version.h    | 2 +-
 2 files changed, 6 insertions(+), 1 deletion(-)

diff --git a/elf/dl-tls.c b/elf/dl-tls.c
index 9fa62f5d..d8f2f740 100644
--- a/elf/dl-tls.c
+++ b/elf/dl-tls.c
@@ -213,6 +213,11 @@ void
 _dl_determine_tlsoffset (void)
 {
   size_t max_align = TLS_TCB_ALIGN;
+  /* libwidevinecdm.so: since 4.10.2252.0 has TLS with 64-byte alignment.
+     Since TLS is initialized before audit modules are loaded and slotinfo
+     information is available, this is not taken into account below in
+     the audit case.  */
+  max_align = MAX (max_align, 64U);
   size_t freetop = 0;
   size_t freebottom = 0;
 
diff --git a/version.h b/version.h
index 83cd1967..2cc5a35a 100644
--- a/version.h
+++ b/version.h
@@ -1,4 +1,4 @@
 /* This file just defines the current version number of libc.  */
 
 #define RELEASE "release"
-#define VERSION "2.32"
+#define VERSION "2.32-arm64tls"
-- 
2.30.0
```

I know it will need a rebase on every glibc package version bump but it's fixed within seconds.
In my opinion in future a version string check will be needed anyway when (ever) glibc proper fix the `dlopen` process and when this patch become obsolete.

dtechsrv:
In fact, it is not the most beautiful solution, but if it works, it's fine.

I already implemeted it to my repo (LE):
[https://github.com/dtechsrv/LibreELEC-AML/commit/4005a2275ef29bc67ce50c783396a443c8e55bf9?branch=4005a2275ef29bc67ce50c783396a443c8e55bf9&diff=unified](https://github.com/dtechsrv/LibreELEC-AML/commit/4005a2275ef29bc67ce50c783396a443c8e55bf9?branch=4005a2275ef29bc67ce50c783396a443c8e55bf9&diff=unified)

PR #443, opened 2021-05-27
Improve TCMalloc detection

mediaminister:
This checks if TCMalloc is preloaded or linked using `/proc/self/maps`

### DIFF
--- a/lib/inputstreamhelper/widevine/arm.py
+++ b/lib/inputstreamhelper/widevine/arm.py
@@ -137,10 +137,13 @@ def install_widevine_arm(backup_path):
     # Google will remove support for older Widevine CDM's on May 31, 2021
     # More info at https://github.com/xbmc/inputstream.adaptive/issues/678 and https://www.widevine.com/news
 
-    # Experimental: Check for TCMalloc support
-    tcmalloc_path = '/usr/lib/libtcmalloc_minimal.so'
+    # Experimental: Check if TCMalloc library exists and gets preloaded
+    libtcmalloc = 'libtcmalloc_minimal.so'
+    process_maps = open('/proc/self/maps', 'r').read()
+    is_tcmalloc_preloaded = bool(libtcmalloc in process_maps)
+
     arm_device = None
-    if not exists(tcmalloc_path):
+    if not is_tcmalloc_preloaded:
         # Propose user to install older version
         if yesno_dialog(localize(30066), localize(30067, os=kodi_os())):  # Your operating system probably doesn't support the newest Widevine CDM. Try older one?
             # Install hardcoded ChromeOS image
--- a/lib/inputstreamhelper/widevine/arm.py
+++ b/lib/inputstreamhelper/widevine/arm.py
@@ -137,7 +137,7 @@ def install_widevine_arm(backup_path):
     # Google will remove support for older Widevine CDM's on May 31, 2021
     # More info at https://github.com/xbmc/inputstream.adaptive/issues/678 and https://www.widevine.com/news
 
-    # Experimental: Check if TCMalloc library exists and gets preloaded
+    # Experimental: Check if TCMalloc library is preloaded or linked
     libtcmalloc = 'libtcmalloc'
     process_maps = open('/proc/self/maps', 'r').read()
     is_tcmalloc_preloaded = bool(libtcmalloc in process_maps)
--- a/lib/inputstreamhelper/widevine/arm.py
+++ b/lib/inputstreamhelper/widevine/arm.py
@@ -138,7 +138,7 @@ def install_widevine_arm(backup_path):
     # More info at https://github.com/xbmc/inputstream.adaptive/issues/678 and https://www.widevine.com/news
 
     # Experimental: Check if TCMalloc library exists and gets preloaded
-    libtcmalloc = 'libtcmalloc_minimal.so'
+    libtcmalloc = 'libtcmalloc'
     process_maps = open('/proc/self/maps', 'r').read()
     is_tcmalloc_preloaded = bool(libtcmalloc in process_maps)
 


dagwieers:
So I wasn't sure if all distributions are using the exact same implementation. But if preloading is the only way to make it work, and it is done when Kodi starts, it means our add-on has libtcmalloc preloaded as well.

So I was wondering within python if we can detect whether libtcmalloc was loaded, or whether we can detect that from our process memory (i.e. /proc/pid).

It would make the implementation somewhat more robust (if my assumptions are correct).

dagwieers:
So I think this can be simplified by doing:

```python
content = open('/proc/self/maps', 'r').read()
```

You do not have to call a unix command to read a file. And we can use the python child process to find if the library is preloaded.

dagwieers:
What would be even better is if we can verify if the calls that Widevine is doing are available. In that case we do not even need to check for the library in the process maps, we could simply ask python if those symbols are available in its process space and conclude they are also available to libwidevinecdm.so.

Any of these would probably do:
```
tcmalloc.pageheap_decommit_count
tcmalloc.pageheap_commit_count
tcmalloc.pageheap_reserve_count
tcmalloc.pageheap_scavenge_count
tcmalloc.aggressive_memory_decommit
tcmalloc: could not allocate bucket
tcmalloc.pageheap_total_decommit_bytes
tcmalloc.pageheap_total_commit_bytes
tcmalloc.slack_bytes
tcmalloc.pageheap_total_reserve_bytes
tcmalloc.max_total_thread_cache_bytes
tcmalloc.current_total_thread_cache_bytes
tcmalloc.pageheap_free_bytes
tcmalloc.transfer_cache_free_bytes
tcmalloc.central_cache_free_bytes
tcmalloc.thread_cache_free_bytes
tcmalloc.sampling_period_bytes
tcmalloc.pageheap_committed_bytes
tcmalloc.pageheap_unmapped_bytes
tcmalloc.central
tcmalloc.transfer
tcmalloc.thread
tcmalloc.page
tcmalloc.page_unmapped
tcmalloc.large
tcmalloc.large_unmapped
```

With ctypes you can easily load a library and then check if any of these symbols are present (and even call them). But I cannot easily find a way to access already loaded (preloaded or linked to python interpreter itself). libraries. It must be possible though.

dagwieers:
But we can make those improvements later as well. It is more important that we have something that works well enough on all supported platforms. Whatever works the broadest is best.

mediaminister:
I'm not sure if `/proc/self/maps` linkt to the right pid because our python code is executed from a Python process.

mediaminister:
Thanks, I simplified the check.

dagwieers:
So, if CoreELEC is linking to libtcmalloc.so or libtcmalloc_minimal.so and it is not preloaded, I assumed that we could no longer check python's process memory for tcmalloc symbols, and this would be a dead-end. I also assumed that using /proc/self/maps will effectively no longer work. But that is not true.

I expected that the parent pid (ppid) of python would be the Kodi binary (kodi.bin) but that is not the case. On LibreELEC the parent of the python interpreter is in fact kodi.sh (the startup wrapper). So the python interpreter is in fact compiled into kodi.bin (big surprise).

In fact that is not a big surprise, because that's how it can cache previous runs of an add-on. And we already established that add-ons were being runs as `LanguageInvoker`. So here you go, the current implementation works fine on CoreELEC just fine. We did a good job, well done ! 😆

lib/inputstreamhelper/widevine/arm.py
```diff
@@ -137,10 +137,13 @@ def install_widevine_arm(backup_path):
     # Google will remove support for older Widevine CDM's on May 31, 2021
     # More info at https://github.com/xbmc/inputstream.adaptive/issues/678 and https://www.widevine.com/news
 
-    # Experimental: Check for TCMalloc support
-    tcmalloc_path = '/usr/lib/libtcmalloc_minimal.so'
+    # Experimental: Check if TCMalloc library exists and gets preloaded
+    libtcmalloc = 'libtcmalloc_minimal.so'
```
dagwieers, line 141: Since CoreELEC allows to link to libtcmalloc too, I would not make this too strict. Chances that some other (non-tcmalloc) library includes `tcmalloc` is pretty improbable.
```suggestion
    libtcmalloc = 'libtcmalloc'
```

dagwieers (changes requested):

lib/inputstreamhelper/widevine/arm.py
```diff
@@ -137,10 +137,13 @@ def install_widevine_arm(backup_path):
     # Google will remove support for older Widevine CDM's on May 31, 2021
     # More info at https://github.com/xbmc/inputstream.adaptive/issues/678 and https://www.widevine.com/news
 
-    # Experimental: Check for TCMalloc support
-    tcmalloc_path = '/usr/lib/libtcmalloc_minimal.so'
+    # Experimental: Check if TCMalloc library exists and gets preloaded
```
dagwieers, line 140:
```suggestion
    # Experimental: Check if TCMalloc library is preloaded or linked
```

dagwieers (changes requested):

dagwieers:
Do we know who can test this on CoreELEC for us?

mediaminister:
I think @CastagnaIT tested the CoreELEC nightly before.

CastagnaIT:
i have tested this PR by choosing "Re-install widevine cdm lib" setting menu
and have reinstalled the cdm without asking to install the old wv library
then i think this works good

mediaminister:
Thanks for confirming.

Portisch:
Thank you for taking care about the linked in option :+1: 

We currently see some users reporting rebooting devices and we maybe switch to the LD_PRELOAD method as it looks "more" stable and causes less rebooting devices.

Currently we trigger two errors with the new cdm lib: rebooting while watching, like after 20-60min and a crash when  a stream is opened/closed multiple times.

CastagnaIT:
@Portisch you are talking about coreelec or libreelec?
you should report the error crash log to https://github.com/xbmc/inputstream.adaptive/issues/678

CastagnaIT:
Tomorrow i will try to play multiple times with my coreelec to see if happen also to me

PR #446, opened 2021-06-01
Don't try installing revoked Widevine CDM

mediaminister:
It makes no sense trying to install the older 4.10.1679.0 Widevine anymore since Google revoked it.


horstle:
I guess this can be merged now.

horstle:
Sonarcloud reported a code smell, which you probably overlooked.

Also, I think the warning message could be improved by mentioning that our detection mechanism might not be reliable due to technical limitations.

mediaminister:
I'll take another look. The length of the warning message should be short enough to fit the ok dialog.
Suggestions for a better warning message are welcome.

PR #450, opened 2021-06-16
Detect TLS 64-bytes support using libc version

mediaminister:


horstle:
We should also keep the tcmalloc check for LE9.2.7. It can be removed at some point, but not now.

mediaminister:
Okay, I added this check again.

horstle:
I think `platform.libc_ver()` should also work? That way we could avoid using `run_cmd()`.

Portisch:
Since tomorrow, 20210621, the string `arm64tls` will be included in CoreELEC 19-Matrix and CoreELEC 20-Nexus nightly and future release builds.

vpeter4:
> I think `platform.libc_ver()` should also work?

Indeed it works:
```
vim3: # python -c 'import platform; print(platform.libc_ver())'
('glibc', '2.32-arm64tls')
```

mediaminister:
@dagwieers @vpeter4 @boot2k3 @dtechsrv @Portisch @horstle Please test with these zip packages and report if this works:
Kodi 19 Matrix: [script.module.inputstreamhelper-0.5.5+matrix.1-tls-1b6a5eb.zip](https://github.com/emilsvennesson/script.module.inputstreamhelper/files/6694004/script.module.inputstreamhelper-0.5.5%2Bmatrix.1-tls-1b6a5eb.zip)
Kodi 18 Leia: [script.module.inputstreamhelper-0.5.5-tls-1b6a5eb.zip](https://github.com/emilsvennesson/script.module.inputstreamhelper/files/6694005/script.module.inputstreamhelper-0.5.5-tls-1b6a5eb.zip)

vpeter4:
I installed version from above on CoreELEC with glibc tls64bytes support patch and it installs widevine cdm 4.10.2255.0. Looks ok to me?

mediaminister:
Yes, without tls64bytes support it shows a warning and installs 4.10.1679.0

vpeter4:
True. I set _has_tls64bytes_support_ to false and I got a warning.

dtechsrv:
It isn't working on LE9.2:
`LibreELEC:~ # ldd --version
ldd (GNU libc) 2.29-arm64tls
$Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
$Written by Roland McGrath and Ulrich Drepper.
LibreELEC:~ # python -c 'import platform; print(platform.libc_ver())'
('glibc', '2.4')
`

Portisch:
Maybe another way is needed to be defined to identify the arm64tls patch

mediaminister:
Running `ldd --version` is the safest way, I guess. I changed this again in https://github.com/emilsvennesson/script.module.inputstreamhelper/pull/455
Please test:
[script.module.inputstreamhelper-0.5.5+matrix.1-tls-4ca54e5.zip](https://github.com/emilsvennesson/script.module.inputstreamhelper/files/6696073/script.module.inputstreamhelper-0.5.5%2Bmatrix.1-tls-4ca54e5.zip)
[script.module.inputstreamhelper-0.5.5-tls-4ca54e5.zip](https://github.com/emilsvennesson/script.module.inputstreamhelper/files/6696074/script.module.inputstreamhelper-0.5.5-tls-4ca54e5.zip)

dtechsrv:
Now it's okay, thank you.

ish_le92_arm64tls

CeEs-atd:
Hello guys,
could anyone please include the arm64tls version string patch for glibc also to LibreElec? This is the latest nightly output...

```
LibreELEC (community): nightly-20210622-48fe299 (H3.arm)
LibreELEC-O:~ # ldd --version
ldd (GNU libc) 2.32
$Copyright (C) 2020 Free Software Foundation, Inc.
```

knaerzche:
This makes you compatible with exactly 1 distro and kidding all other users.  You may want to check` /etc/os-release` what the actual distro is.

vpeter4:
> You may want to check `/etc/os-release` what the actual distro is

Not just distro but also it's version.

Portisch:
> This makes you compatible with exactly 1 distro and kidding all other users. You may want to check` /etc/os-release` what the actual distro is.

Why LibreELEC just pop up AFTER a solution was found? Tell us such ideas 2 weeks ago would be helpful.

But if nobody make a PR to LE it would never be recognized, my opinion.

knaerzche:
> Why LibreELEC just pop up AFTER a solution was found? Tell us such ideas 2 weeks ago would be helpful.

Well, I wouldn't call that a solution: In general software really should not expect an OS to do unnecessary changes to core components for such a trivial thing, which could be solved in so many other ways.

I gave a pointer how that could be solved with a pretty easy change (which took me 3 secs to think about) which works on virtually any linux distro. Its completly up to you what you make out of it (or not).

Currently this addon provides false information and it'd be an better addon without having this "detection" at all.  In fact: its a bug.

dtechsrv:
This "detection" had to be included in the addon due to the usability of older (non-patched) systems.
(But, it’s true that we don’t know at all how long these will work with old Widevine.)

> In fact: its a bug.

Okay, but keep in mind that this error was generated by Google.

knaerzche:
> This "detection" had to be included in the addon due to the usability of older (non-patched) systems.
> (But, it’s true that we don’t know at all how long these will work with old Widevine.)

Sure, I got that. I just don't think its done in a proper way. Please re-read.

> Okay, but keep in mind that this error was generated by Google.

I wouldn't blame google for that. We are abusing a binary made for a different operating system here so this was expected to happen at some point.

horstle:
@knaerzche Your solution would mean we had to create and maintain a list with compatible systems. If another system becomes compatible, an update to that list would require a new version of the addon, until that's released and updated by the user, the addon would again provide wrong information.

On the other hand, any system which wants to be compatible with newer Widevine releases has to apply two patches anyway. Since we can't easily detect if these are applied, adding a detectable tiny third patch which is barely any additional effort seems reasonable to me.

Of course, once the older Widevine versions are revoked, this whole detection mechanism becomes much less relevant, since then we can only tell the user "your system is incompatible" which he will find out a few minutes later anyway.

CeEs-atd:
When I understood right, the TLS64 and RELR patches itself are not the problem for libreELEC, it's only the libc version string change?
What are the options:

- keeping a list of "known working distros & patchlevels" inside the ISH code or sowewhere online is also not a good solution (maintenance, availability of webservice, .)
- having a ARMTLS64 entry in  `/etc/os-release` of every distro... which can be read by the script? This would end up in an ability lineup, if every scripts wants to assure specific OS abilites/patches/
- checking the OS TLS64 abilitiy directly with a kind of test lib is also not wanted (due to unwanted binary distribution or needless download of chrome OS images),
- getting error codes returned from IS is also not in short reach...

So why not add a persistent "ARMTLS64 present" switch in ISH settings? Default is "on" This would keep existing and working installations with patched OS usable without automated tests. For users who experience trouble with the CDM the switch can be set to "off", which could result in the "do you want to try an older WV version?" dialogue. 
Having a working manual changeable configuration solution is better than a sometimes not working automated test...?

(An additional "Test OS features & DRM compatibilty" - Button could try an automated test and make some suggestions to the user..?)

knaerzche:
>  Your solution would mean we had to create and maintain a list with compatible systems. 
Yes, you shoud do that if you want to provide any information which is correct. I wouldn't expect any bigger distro to include those patches.

horstle:
> if you want to provide any information which is correct

As I said in the sentence right after that, the information wouldn't be any more correct!


> until that's released and updated by the user, the addon would again provide wrong information.


About the second sentence:
> I wouldn't expect any bigger distro to include those patches.

I don't think that's too unrealistic, but even if it is, my argument still stands for fewer smaller distros. It might be even worse there, since we might not know about them becoming compatible at all.

Would you maybe be more comfortable with the version string change being part of the second patch?

knaerzche:
> As I said in the sentence right after that, the information wouldn't be any more correct!

Its not correct already: So where's your point? (btw: check your tone)

> Would you maybe be more comfortable with the version string change being part of the second patch?

I can't speak for LE as a whole, but I don't think this version string change will ever get included: glibc is a core component of the operating system and the version string might be checked by other packages which might fail after that change.

horstle:
My point was, that your change wouldn't be an improvement, since you mentioned the "false info" as an issue multiple times. By the way I don't think my tone was any worse than yours (except maybe that I used an exclamation mark, I removed that now), but all that is moot, since


> the version string might be checked by other packages which might fail after that change.

this is actually an issue I agree with.
Although it would have been great if you had mentioned that earlier, I agree now, that we should find a different solution. If no better solution comes up, as a last resort the user might just have to download the Chrome OS image so we can check compatibility directly.

dtechsrv:
Another aproach:
If the glibc version string is not acceptable, even an environment variable could be set in the system, and if it exists, it can already be used to check compatibility.

For example:
LIBWIDEVINECDM_PATCH_VERSION=4.10.2252.0
... or something like this.

With the version number, you would be able to check the patches later if Google makes a change again.
Environment variables exist under all OSs, and so it’s not that painful because it doesn’t involve changing any packages.

knaerzche:
> My point was, that your change wouldn't be an improvement, since you mentioned the "false info" 

If you're afraid changes in that particular case happen to fast for addon updates: provide a arm64tls.json at github that lists compatible OS'es which the addon can check against .... or something like that: be creative ;)

Portisch:
but I don't think this version string change will ever get included
So we all need to make it "better" as LE will not accept the changes? So the user world needs to follow LE, not vice versa...

    and the version string might be checked by other packages which might fail after that change
This look again for me LE is not at least testing or make any proof of concept when changes are not from LE side.
Because then you would have seen that there is no side effect - yet.

I am really sorry to released any patches/solution about this issue. I didn't have done the PR to LE, so you were never forced to merge it from my side. And all comments from LE side shows it's a unpopular patch anyway. So please make me a pleasure and drop the glibc patch at all in LE that we can stop this discussion, thank you.

dtechsrv:
> If you're afraid changes in that particular case happen to fast for addon updates: provide a arm64tls.json at github that lists compatible OS'es which the addon can check against .... or something like that: be creative ;)

Please don't sit on the horse backwards!
Don’t force the addon builder to collect which ARM OS is supported because there are a couple of them.

horstle:
@dtechsrv Using an environment variable would be fine by me. It should be very easy to implement this in ISH. What do the others think about that? Is that easy to do in CE/LE? Would it also be doable for maintainers of the kodi package on desktop distros?

CvH:
To keep it realistic no bigger distribution will add a hack to glibc (I am not gone see that happen for LE ) to allow an easy read out for some single kodi helper addon.

Collecting a working OS/Version list sounds also a bit overkill. 

Adding an ENV is a better choice but maybe a much easier solution would be checking if some file with some values exist at the inputstream addon itself? So just some minor patch would be required to ensure compatibility is there or not.
Depending of the Distro a bump of the repository may be needed but this is a minor thing (the glibc patch is needed anyway).
Tbh I am not seeing any OS besides *ELEC adapting the glibc patch anyway.
Maybe I am overlooking something here - not sure.

Maybe starting with july 2021 all these stuff is obsolete, who knows.

vpeter4:
This works btw:
```
$ git diff
diff --git a/projects/Amlogic-ce/packages/mediacenter/kodi/config/kodi.conf b/projects/Amlogic-ce/packages/mediacenter/kodi/config/kodi.conf
index fdb943b8f4..fcf1b8fb7e 100644
--- a/projects/Amlogic-ce/packages/mediacenter/kodi/config/kodi.conf
+++ b/projects/Amlogic-ce/packages/mediacenter/kodi/config/kodi.conf
@@ -2,3 +2,4 @@ KODI_AE_SINK=ALSA+PULSE
 HOME=/storage
 KODI_TEMP=/storage/.kodi/temp
 KODI_HOME=/usr/share/kodi/
+ARM64TLS=yes
```

`16:57:27.291     INFO <general>: environment variable ARM64TLS: yes`

CvH:
maybe adding `ARM64TLS=1.0` instead ? 
if some stuff in future requires additional fixes it could easily checked against `$var >= 1.0` or something

HiassofT:
I also think checking an environment variable is a quite good solution.

As for naming: how about `LIBC_WIDEVINE_PATCHLEVEL=1.1` or sth like that? That'd be more generic in case you need additional patches in the future.

horstle:
Since it seems we can all agree on that (thanks again to @dtechsrv ), I created a PR. Name and value of the env variable could of course be changed if wanted.

dtechsrv:
I think an integer is more beautiful than a float (no point is required), and since this is the first version, we can start from 1.
Later 2, 3, etc. Anyway, it will be good. ;-)

dtechsrv:
The version number is technically not a float, so the statement is not valid. (e.g. 1.2 > 1.11 and 1.1 == 1.10)
Version numbers should be handled separately (major.minor), but actually an integer is enough for this variable.
And since this is the first version, we can start with number 1. (Later 2, 3, etc.)

But, it looks good anyway. ;-)

This patch is also working on LE9.2 and LE10 (with this modified variable):
```
diff --git a/packages/mediacenter/kodi/scripts/kodi-config b/packages/mediacenter/kodi/scripts/kodi-config
index 5844ebc6c0..e86489d696 100755
--- a/packages/mediacenter/kodi/scripts/kodi-config
+++ b/packages/mediacenter/kodi/scripts/kodi-config
@@ -43,6 +43,8 @@ else #arm
   echo "MALLOC_MMAP_THRESHOLD_=8192" >> /run/libreelec/kodi.conf
 fi

+echo "LIBC_WIDEVINE_PATCHLEVEL=1" >> /run/libreelec/kodi.conf
+
 if [ -f /storage/.config/kodi.conf ] ; then
   cat /storage/.config/kodi.conf >>/run/libreelec/kodi.conf
 fi
```

vpeter4:
I think more appropriate place would be directly in `packages/mediacenter/kodi/config/kodi.conf`. No need to dynamically add this option to config file.

dtechsrv:
There is no "packages/mediacenter/kodi/config/kodi.conf" in LE9.2 (or CE9.2).

horstle:
Changed it to int.

vpeter4:
True, then use different files for different versions.

dtechsrv:
> I think more appropriate place would be directly in `packages/mediacenter/kodi/config/kodi.conf`. No need to dynamically add this option to config file.

I've been thinking, and the static solution isn't good from the OS side after all.
Under CoreELEC, it doesn't really matter, but LibreELEC is also made for x86_64.
But under non-arm, this patch is not required to the best of my knowledge, so this variable is not needed either.

We need a check to set this variable to kodi.conf only under arm* (arm/aarch64/arm64) architecture.
That can be good?
```
if [ "$(uname -m)" = "arm" ] || [ "$(uname -m)" = "aarch64" ] || [ "$(uname -m)" = "arm64" ]; then
  echo "LIBC_WIDEVINE_PATCHLEVEL=1" >> /run/libreelec/kodi.conf
fi
```

vpeter4:
Maybe better to use `-n`? And why a difference between getting arm or aarch64?
`if [ -n "$(uname -m | grep arm)" ] || [ "$(uname -m)" == "aarch64" ]; then`

dtechsrv:
Okay, with `-n` it looks better.
> And why a difference between getting arm or aarch64?

The arm* may even contain something else in place of the wildcard, but aarch64 is just aarch64, to my knowledge there can be nothing else.

PR #452, opened 2021-06-18
Postpone update by 2 days if declined by user

### LINKED ISSUE #451: Postpone reasking Widevine update with 24h or 48h
Currently ISH will ask the user to update Widevine CDM on every playback request, which is painful if you reject it because the timing is not right or you have other reason for not doing it.

I think we are annoying our users needlessly with this. Therefore I would propose we postpone asking the user again in 24h or 48h if he rejects it.

Obviously #69 would be better, but may not be possible before the next major Kodi release (the next few years).

horstle:
This fixes #451 .

Should we maybe make the time configurable? I don't think it's really necessary.

### DIFF
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -263,6 +263,11 @@ class Helper:
         """Prompts user to upgrade Widevine CDM when a newer version is available."""
         from time import localtime, strftime, time
 
+        update_declined_at = get_setting_float('update_declined_at', 0.0)
+        if update_declined_at + 3600 * 24 * 2 >= time():
+            log(2, 'User had declined an update on {date}', date=strftime('%Y-%m-%d %H:%M', localtime(update_declined_at)))
+            return
+
         last_check = get_setting_float('last_check', 0.0)
         if last_check and not self._first_run():
             if last_check + 3600 * 24 * get_setting_int('update_frequency', 14) >= time():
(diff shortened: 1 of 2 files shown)


dagwieers:
I would not make it configurable, but we could debate the pros and cons of the various options. 1d, 2d, 3d or 1w ?

There are different situations to consider.

mediaminister:
I guess this is not needed because the default update frequency is 31 days and in https://github.com/emilsvennesson/script.module.inputstreamhelper/pull/450 I changed the way `last_check` is set.
`last_check` is now set from the moment the version check is done, regardless of what the user decided or what version is installed.

So, I guess the user will only asked once a month to update.

dagwieers:
@mediaminister Owh, I missed that because I am running official releases. However, only asking once ever 31 months is probably not a good option either.

If people on playback delay updating Widevine CDM (we all do this on ARM, I just want to watch something _now_, not wait 10 minutes with the family), and Widevine is revoked, your setup will fail and you may not remember it could be because you canceled the update prompt 2 weeks before.

On ARM what we could do is match the version with the x64 releases, and only ask for an update if e.g. x64 has an actual new release. (it is a cheap check anyway, although we have to cache those releases as well)

This way we can ask to update only when a new release is expected (e.g. ARM leaps behind) but not ask if we are still ahead (except every 31 days).

mediaminister:
> and Widevine is revoked, your setup will fail and you may not remember it could be because you canceled the update prompt 2 weeks before.

I understand, I'll guess I should first solve a problem in the TLS 64-bytes support PR when an older ARM version is installed.

PR #453, opened 2021-06-22
Always set last_check timestamp

mediaminister:
This makes sure the `update_frequency` setting is respected regardless of which version the user has installed. This prevents that the user is annoyed with unwanted update prompts.


mediaminister:
I guess this is not the solution we want.

PR #458, opened 2021-07-02
Prepare for v0.5.7 release

horstle:


horstle:
Why is pylint compaining about that now?

horstle:
Should we maybe remove `install_widevine_arm_legacy` this release?

dtechsrv:
>Why is pylint compaining about that now?

Unfortunately, I don't understand the python language, but it's interesting that it gives an error to a line that hasn't been modified since the last couple of releases.

> Should we maybe remove `install_widevine_arm_legacy` this release?

Is this still some old method dating back to "pre-CromeOS" images?

horstle:
> Is this still some old method dating back to "pre-CromeOS" images?

No, it's an old method to extract the widevine library from the ChromeOS image which required the image to be extracted completely.

I think I'll just quickly fix the pylint error. Removing that function would also mean removing multiple functions which are only required by that old method. I'd rather test that properly to make sure I didn't remove too much.

horstle:
I'll release this tomorrow, if there are no objections @dagwieers, @mediaminister.

dtechsrv:
All right, thanks a lot. ;-)

mediaminister:
No objections. Looks good to me.

PR #461, opened 2021-07-20
Remove deprecated install method on ARM.

horstle:
This removes the deprecated method to extract Widevine from the Chrome OS image and everything associated with it.

I'm not sure about the strings.po files. i'll sync the other languages to the english file in a seperate PR, but is it ok, if some IDs in the middle are missing? Changing all the IDs of the following messages would be a rather big task which seems unnecessary to me.


horstle:
https://cros-updates-serving.appspot.com/csv is apparently not reachable atm. All the other checks passed.

mediaminister:
> but is it ok, if some IDs in the middle are missing?

Okay for me. You can merge this.

PR #468, opened 2021-09-09
Remove TLS 64-byte alignment detection

mediaminister:
There is only one working Widevine CDM for ARM systems today and it needs a system that supports TLS 64-byte alignment. It doesn't make a lot of sense for ISH to detect if the users system is compatible. If not , Widevine will definitely fail and users should upgrade their system anyway.


horstle:
Technically we don't need the `kodi_os()` function any more, but I wouldn't mind keeping it. Other then that, looks good. IMO you can merge and I think it's best to immediately do a release afterwards.

PR #481, opened 2022-02-05
avoid distutils usage

MarkusVolk:
LooseVersion is part of distutils which will be removed from python 3.12.
Add a simplified version found in sympy and use that instead.

Signed-off-by: Markus Volk <f_l_k@t-online.de>


mediaminister:
Thanks, because I couldn't merge this from your master branch I had to move this to https://github.com/emilsvennesson/script.module.inputstreamhelper/pull/484

PR #484, opened 2022-02-06
avoid distutils usage

mediaminister:
LooseVersion is part of distutils which will be removed from python 3.12.
Add a simplified version found in sympy and use that instead.

Signed-off-by: Markus Volk <f_l_k@t-online.de>


mediaminister:
Thanks for the heads-up.
But I rather will replace `LooseVersion` with `packaging.version.parse`

Issue #495, opened 2022-03-11
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 58437: ordinal not in range(128)


DanielGabris:
Hello, 

since today I get following error when trying to start any movie on netflix addon. 
"InputStream Helper has geberated an internal error:: 'asci' codec cat't decode byte 0xe2 in position 58437: ordinal not in range (128)
KODI version 18.7 runninig on Linux arm (raspberry 3)
InputStream Helpre is at version 0.5.8

also get error when tried to reinstall widevine CDM.
 Worked well until today, nothing was updated lately. 

error from log:
2022-03-11 17:57:46.346 T:1354748096   ERROR: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                             - NOTE: IGNORING THIS CAN LEAD TO MEMORY LEAKS!
                                            Error Type: <type 'exceptions.UnicodeDecodeError'>
                                            Error Contents: 'ascii' codec can't decode byte 0xe2 in position 58437: ordinal not in range(128)
                                            Traceback (most recent call last):
                                              File "/home/pi/.kodi/addons/script.module.inputstreamhelper/default.py", line 8, in <module>
                                                run(sys.argv)
                                              File "/home/pi/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/api.py", line 14, in run
                                                widevine_install()
                                              File "/home/pi/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/api.py", line 42, in widevine_install
                                                Helper('mpd', drm='widevine').install_widevine()
                                              File "/home/pi/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/__init__.py", line 34, in clean_before_after
                                                result = func(self, *args, **kwargs)
                                              File "/home/pi/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/__init__.py", line 221, in install_widevine
                                                result = install_widevine_arm(backup_path())
                                              File "/home/pi/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/widevine/arm.py", line 69, in install_widevine_arm
                                                devices = chromeos_config()
                                              File "/home/pi/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/widevine/arm.py", line 63, in chromeos_config
[...]

PR #496, opened 2022-03-12
Interpret http-get feedback as utf-8 (issue #495)

archtur:
I tried the [proposal from DokuroKM](https://github.com/emilsvennesson/script.module.inputstreamhelper/issues/495#issuecomment-1065477688) in issue #495, and this helped. 
I think the root cause is, that google changed the format of the information which the input stream helper uses to check the latest available widevine version.

On my system after the change of the local file `~/.kodi/addons/script.module.inputstreamhelper/lib/inputstreamhelper/utils.py` the widevine update dialog was shown correctly and I was able to update. After the update everything worked as intented.

Can please someone review that and check that it has no side effects?

### DIFF
--- a/lib/inputstreamhelper/utils.py
+++ b/lib/inputstreamhelper/utils.py
@@ -70,7 +70,7 @@ def http_get(url):
     content = req.read()
     # NOTE: Do not log reponse (as could be large)
     # log(0, 'Response: {response}', response=content)
-    return content.decode()
+    return content.decode("utf-8")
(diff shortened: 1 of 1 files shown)


horstle:
From my testing this change should not affect other calls of http_get and also seems to work well on Python 3. If the tests finish fine, I'll merge this tomorrow and probably do a release right after.

horstle:
This will fix #495

PR #506, opened 2022-03-23
Update german translation for version 0.5.9

tweimer:


mediaminister:
Thanks for the translation

PR #517, opened 2022-04-18
Prepare for v0.5.10 release

horstle:


horstle:
If there are no objections, I'll publish this release this evening.

PR #533, opened 2023-03-24
Initial support of 64bit userspace on ARM

### LINKED ISSUE #530: Widevine AARCH64 now available
64bit ARM Linux Widevine CDM now out in the wild :slightly_smiling_face:
@tmm1 found it. looks like was just released last month in a new 64bit beta chrome os

"
seems new development https://groups.google.com/a/chromium.org/g/chromium-os-dev/c/GdZ0mebutXw
inside https://gsdview.appspot.com/chromeos-localmirror/distfiles/chromeos-lacros-arm64-squash-zstd-112.0.5579.0
"

i managed to open the above squash-zstd file in windows using an extended version of 7zip: https://github.com/mcmilk/7-Zip-zstd

tmm1 said its not working in the arm64 build of LibreELEC due to "libwidevinecdm.so: undefined symbol: __aarch64_ldadd4_acq_rel" but the LibreELEC guys are looking into it

inputstream helper will need to add support for this.
Not sure if there is any recovery image that contains it yet.
Unless there is a Beta channel recovery.
Looking for Acer Chromebook Spin 13 (CP713-1WN)  on Chrome version 112
Current "Stable" seems to be chrome 109

UPDATE:
"LE folks got aarch64 widevine working! Might require some glibc patches, will see how it shakes out"

horstle:
Closes #530 

So far only LAZOR (and some others which use the same image) has the 64bit library. We need to monitor the other images for when/if they make the switch.
The path for `find_file()` could be changed accordingly, since the lib is not in `arm_cros`, but `arm_cros64` in those images.

### DIFF
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -9,7 +9,7 @@ from . import config
 from .kodiutils import (addon_version, browsesingle, delete, exists, get_proxies, get_setting, get_setting_bool, get_setting_float, get_setting_int, jsonrpc,
                         kodi_to_ascii, kodi_version, listdir, localize, log, notification, ok_dialog, progress_dialog, select_dialog,
                         set_setting, set_setting_bool, textviewer, translate_path, yesno_dialog)
-from .utils import arch, http_download, parse_version, remove_tree, store, system_os, temp_path, unzip
+from .utils import arch, http_download, parse_version, remove_tree, store, system_os, temp_path, unzip, userspace64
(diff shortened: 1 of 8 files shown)


horstle:
If anybody want's to test this, please install 
[script.module.inputstreamhelper-0.5.10+matrix.1-arm64-3983f12.zip](https://github.com/emilsvennesson/script.module.inputstreamhelper/files/11072061/script.module.inputstreamhelper-0.5.10%2Bmatrix.1-arm64-3983f12.zip)

horstle:
I tested it as much as I could, and it seems fine. Going to merge it now.

Issue #421, opened 2021-01-09
Temporary directory not saved until you exit the addon


GutHib:
Here's a bug that drove me absolutely nuts today... hope it will save others some headache: Specifying a new temporary download directory and hitting OK will not actually configure&save that directory, until you hit another OK button and leave the addon.

It seemed like all my storage devices were too small or broken or poorly formatted or whatever else. Nope: The addon didn't even look at them...

horstle:
Yes, we discussed this somewhere before. In Kodi, settings are only stored after you close the settings page. Not sure what we could do about it.

GutHib:
You could leave a note on the page where you specify the temporary download directory.

Preferably in red. ;)

PR #538, opened 2023-04-25
Create option to choose Widevine version to install from repo.

### LINKED ISSUE #537: Feature requests for more expert control
I am currently involved in the ISAdaptive project that its the only one that support this add-on,
i noticed these limitations on your add-on where it could be more feature-rich
to facilitate development for us in the Kodi Team as well, not only final users

Currently more times i have to do all the operations (install/copy/replace widevine) manually and then download all widewine versions locally and use webbrowser to get new versions etc... always manually or as workaround i have to install an addon that integrate ISH just to run Widevine installation...
so having some improvement on this aspect would be useful

Mainly two requests:

1) Have a way to install CDM from ISH settings, atm its mandatory to have a video add-on that integrate ISH without it your add-on is totally useless. When we developing/debug usually we dont always install video add-ons, so having a way to manually install CDM from expert settings would be very useful to us

2) Have a way to install choosable CDM versions available on the google server, this is not available for all platforms i know, but for supported platforms its must have. The backup functionality is interesting, but totally useless when you have an empty Kodi system

Practically in the expert settings i suggest having:
- Install CDM (allow install CDM without depending from addons, where allow to choose the version, last version pre-selected, this will also replace possible previous installation with selected version) disable this menu to unsupported platforms, at least support common platform of https://dl.google.com/widevine-cdm/versions.txt
- Re-install CDM (unchanged)
- Remove CDM (unchanged)
- Restore CDM (unchanged)

horstle:
Closes #537 

Test zips:
[script.module.inputstreamhelper-0.5.10-install_choosable-9c46011.zip](https://github.com/emilsvennesson/script.module.inputstreamhelper/files/11360932/script.module.inputstreamhelper-0.5.10-install_choosable-9c46011.zip)
[script.module.inputstreamhelper-0.5.10+matrix.1-install_choosable-9c46011.zip](https://github.com/emilsvennesson/script.module.inputstreamhelper/files/11360933/script.module.inputstreamhelper-0.5.10%2Bmatrix.1-install_choosable-9c46011.zip)


CastagnaIT:
tested on a clean Kodi 21 but when i coose re-install addon is crashed
https://paste.kodi.tv/awemenexiy.kodi

horstle:
Your system apparently doesn't have distutils, so the addon uses a deprecated method as fallback, but the variable returned by that is not hashable. Should be fixed now. I also updated the zips in the original comment.

CastagnaIT:
> Your system apparently doesn't have setuptools

on Windows kodi have python embedded dont use system installed python

tested new link above but now raise another error:
https://paste.kodi.tv/gokojoyetu.kodi

horstle:
Right, stupid mistake...

Fixed and new zips in the start post.

CastagnaIT:
yes now works thanks
just a little improvement, after you installed an older CDM version it immediately ask to update to last version
if you have just installed an old version, there is no point in asking me to update to the latest version immediately
imo this operation that should be automatically skipped and delayed on the specified days

horstle:
Actually it does that. After the install finished, you probably didn't us "OK" to exit the settings. The addon stores the install time in a hidden setting, but that doesn't actually get saved, if you don't press "OK". It's basically the same issue as #421.

PR #561, opened 2024-05-13
Implement own "Version" class to get rid of distutils/packaging dependency

### LINKED ISSUE #550: Python 3.12 removing distutils
Python 3.12 will finally remove the long-deprecated distutils. So the fallback for `parse_version()` will no longer work. I don't think we can assume, that `packaging` will be installed everywhere.

horstle:
Fixes #550.

### DIFF
--- a/lib/inputstreamhelper/utils.py
+++ b/lib/inputstreamhelper/utils.py
@@ -8,6 +8,8 @@ from time import time
 from socket import timeout
 from ssl import SSLError
 import struct
+from typing import NamedTuple
+from functools import total_ordering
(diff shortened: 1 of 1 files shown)


jaylinski:
@horstle I just updated to Ubuntu 24.04 and Kodi 21.1 (flatpak). With this configuration, the current release (v0.6.1) of this module is broken because (I think) Python 3.12 is used which removed distutils.

So it would be awesome if you could tag and publish a new release.

(I managed to fix the issue by manully applying this PR.)

PR #573, opened 2024-09-23
prepare for v0.7.0

emilsvennesson:
I decided on 0.7.0 for this version because we are removing support for Python 2.

@horstle, do you think we should include anything else in this version? I'd appreciate your eyes on this seeing as you've done the heavy lifting as usual. :-)


horstle:
Apart from removing the import that pylint is moaning about, I wouldn't add anything else.

Actually I still haven't tested #558 as much as I wanted to before a release (currently don't have the time for it), but I'm rather confident it's fine.

horstle (approved):

emilsvennesson:
@horstle Thanks! I pushed the release now.

PR #578, opened 2024-10-05
pylint fixes

emilsvennesson:


lib/inputstreamhelper/utils.py
```diff
@@ -73,143 +72,168 @@ def download_path(url):
     return os.path.join(temp_path(), filename)
 
 
-def _http_request(url, headers=None, time_out=10):
-    """Perform an HTTP request and return request"""
-    log(0, 'Request URL: {url}', url=url)
-
+def _http_request(url, headers=None, time_out=30):
```
horstle, line 75: the headers parameter would be unused after this change

lib/inputstreamhelper/utils.py
```diff
@@ -73,143 +72,168 @@ def download_path(url):
...
-            raise HTTPError('HTTP {} Error for url: {}'.format(req.getcode(), url), response=req)
+        response = urlopen(url, timeout=time_out)  # pylint: disable=consider-using-with:w
+        if response.status in [301, 302, 303, 307, 308]:  # Handle redirections
+            new_url = response.getheader('Location')
+            log(1, f"Redirecting to {new_url}")
+            return _http_request(new_url, time_out)
+        return response  # Return the response for streaming
     except (HTTPError, URLError) as err:
         log(2, 'Download failed with error {}'.format(err))
-        if yesno_dialog(localize(30004), '{line1}\n{line2}'.format(line1=localize(30063), line2=localize(30065))):  # Internet down, try again?
```
horstle, line 91: Please leave this comment in, so one doesn't have to look up what message 30065 says.

lib/inputstreamhelper/utils.py
```diff
@@ -73,143 +72,168 @@ def download_path(url):
...
-                                                                               line2=localize(30065))):  # Could not finish dl. Try again?
-                    progress.close()
-                    return False
-
-                headers = {'Range': 'bytes={}-{}'.format(size, total_length)}
-                req = _http_request(url, headers=headers)
-                if req is None:
-                    return None
-                continue
+        while True:
```
horstle, line 183: I try to avoid `while True`. To make sure we can never get stuck in an infinite loop, I would stick with `while size < total_length`

lib/inputstreamhelper/utils.py
```diff
@@ -73,143 +72,168 @@ def download_path(url):
...
-                progress.close()
-                req.close()
                 return False
-            if time() - starttime > 5:
+
+            if time() - starttime > 5 and size > 0:
                 time_left = int(round((total_length - size) * (time() - starttime) / size))
                 prog_message = '{line1}\n{line2}'.format(
                     line1=message,
-                    line2=localize(30058, mins=time_left // 60, secs=time_left % 60))  # Time remaining
```
horstle, line 186: Again, keep the comment. I don't want to have to look up what message 30058 is.

lib/inputstreamhelper/widevine/repo.py
```diff
@@ -19,7 +19,10 @@ def cdm_from_repo():
 
 def widevines_available_from_repo():
     """Returns all available Widevine CDM versions and urls from Google's library CDM repository"""
-    cdm_versions = http_get(config.WIDEVINE_VERSIONS_URL).strip('\n').split('\n')
+    cdm_versions = http_get(config.WIDEVINE_VERSIONS_URL)
+    log(0, f"Available Widevine versions from repo: {cdm_versions}")
+    cdm_versions = cdm_versions.strip('\n').split('\n')
+    log(0, f"Available Widevine versions from repo: {cdm_versions}")
```
horstle, line 25: I think this is a bit much, even in a debug log

horstle (commented):
All in all some nice improvements.

PR #582, opened 2024-10-13
Quick fix disabling Lacros on XBox.

horstle:
Since I don't have an XBox I can't test if/how we could use zstd there, so this disables it.


tpaniaki:
Any deadline on this?

